Issue #44, opened 2026-04-29
Ingestor: declarative YAML config schema and generic entrypoint


saadqbal:
Parent: tracebloc/client#85

## What

Replace the per-customer Python script + Dockerfile pattern with a single declarative `ingest.yaml` that the official image consumes. Customers describe *what* they want ingested; the platform owns *how* it happens.

### YAML schema (`schema/ingest.v1.json`)

JSON Schema covering all current task categories:

- image classification, object detection, text classification
- tabular classification, tabular regression
- time-series forecasting, time-to-event prediction

Top-level shape (illustrative, finalize during PR):

\`\`\`yaml
apiVersion: tracebloc.io/v1
kind: IngestConfig
metadata:
  table: claims_train
  intent: train     # train | test
  category: tabular_regression
spec:
  source:
    type: csv
    path: /data/shared/claims/train.csv
    options: { delimiter: \",\", encoding: utf-8 }
  schema:
    age: INT
    income: DOUBLE
    target: DOUBLE
  data_id:
    strategy: uuid           # uuid | column
    # column: claim_id       # required if strategy=column; warns that values leave the cluster
  label:
    column: target
    policy: bucket           # bucket | passthrough  (REQUIRED for regression-class tasks)
    bucket:
      strategy: quantile     # quantile | fixed_width
      bins: 10
  validators:
    - file_exists
    - numeric_columns
  file_options:
    target_size: [512, 512]
    extension: jpg
  customScript: /custom/transform.py   # optional escape hatch
\`\`\`

Schema validation must reject regression-class tasks without an explicit `label.policy`.

### Generic entrypoint (`tracebloc_ingestor/cli/run.py`)

- Reads `INGEST_CONFIG` env (path to YAML).
- Validates against schema; fails fast with line-number errors.
- Dispatches to the right ingestor (CSV/JSON) based on `source.type`.
- Wires validators from the YAML rather than from per-script Python.
- If `customScript` is set, imports the file (mounted via ConfigMap) and exposes its `transform(record)` hook between read and validate.

### Examples

Convert each `templates/*/<name>.py` into an equivalent `examples/yaml/<name>.yaml` so users have a working starting point per task category.

## Acceptance criteria

- An `ingest.yaml` covering each of the seven task categories produces the same end state today's per-script-Python templates produce (same MySQL rows, same backend POSTs).
- Regression-class YAML without `label.policy` fails validation before any DB connection or network call.
[...]

PR #119, opened 2026-05-22
fix: close YAML-vs-template parity gaps in per-category metadata

divyasinghds:
## Summary

The YAML path (#44) silently dropped three pieces of per-category config that the template scripts carried, so YAML-driven runs produced different backend metadata / parsed records than equivalent template runs. Each gap was identified by walking each `templates/*/` file against the YAML resolver.

### 1. Tabular family — `number_of_columns` never sent

Templates set `file_options={\"number_of_columns\": len(schema)}` at the call site. The YAML resolver returns `file_options={}` for tabular, and `BaseIngestor.__init__` only refreshed the value when the key was already present:

```python
if \"number_of_columns\" in self.file_options:           # gated
    self.file_options[\"number_of_columns\"] = len(table_schema)
```

So YAML-path tabular runs never sent `number_of_columns` to the backend. Removed the guard — when a schema is supplied, `number_of_columns` is always set from the **cleaned** schema length (after label / annotation / unique_id columns are stripped).

### 2. keypoint_detection — `target_size` / `number_of_keypoints`

Both are dataset-specific (pose-model input resolution and the dataset's keypoint count); no convention default makes sense. The resolver was returning `target_size=[448, 448]` (template uses 256×256) and never set `number_of_keypoints` at all.

- Schema: new top-level `target_size` and `number_of_keypoints` properties; `if/then` requires both for `category: keypoint_detection`.
- Resolver: drop `target_size` from the keypoint default block; bridge top-level fields into `file_options` (precedence: `spec.file_options` > top-level > category default).
- Example YAML: updated with the required fields and explanatory comments.

### 3. Tabular family — `na_values` cell-parsing divergence

Templates pass `na_values=[\"\", \"NA\", \"NULL\", \"None\"]` via csv_options. `spec.csv_options` in the schema is `additionalProperties: false` and only accepts `{chunk_size, delimiter, quotechar, escapechar}`, so the YAML path stored `\"NA\"` / `\"NULL\"` / `\"None\"` cells **as strings** instead of NaN — a silent data-parsing diff against template runs.

Widened CSVIngestor's internal `na_values` default to the wider list for the four tabular-family categories. Template runs are unaffected: their explicit `csv_options` still wins via the existing `{**default_options, **self.csv_options}` merge.

Out of scope (deliberately, per discussion): `chunk_size` mismatch (template uses 100 for image/text-heavy categories vs 1000 default — perf, not correctness).

## Test plan
- [ ] `pytest tests/test_csv_file_options.py` — YAML-path `file_options={}` now asserts `number_of_columns == len(cleaned_schema)`.
- [ ] `pytest tests/test_conventions.py::test_keypoint_detection_bridges_top_level_fields` — top-level fields land in `file_options`.
- [ ] `pytest tests/test_template_equivalence.py::test_yaml_resolves_to_template_equivalent_kwargs[keypoint_detection]` — 256×256 + 9 keypoints, no stale 448×448 default.
- [ ] `pytest tests/test_csv_na_values.py` — tabular-family categories get wide `na_values`; image_classification keeps narrow.
- [ ] Manual: validate an `ingest.yaml` for `category: keypoint_detection` without `target_size` or `number_of_keypoints` and confirm the entrypoint fails fast with a schema error pointing at the missing fields.

---

> [!NOTE]
> **Medium Risk**
> Touches ingest schema validation and per-category defaulting/CSV parsing, which can change runtime metadata and how tabular values are interpreted. Scope is moderate but well-covered by new/updated unit tests across conventions, equivalence, and CSV parsing.
> 
> **Overview**
> Improves YAML-vs-template parity for per-category metadata and CSV parsing.
> 
> **Keypoint detection** now requires top-level `target_size` and `number_of_keypoints` in the JSON schema and example YAML; the resolver drops the keypoint `target_size` default and bridges these top-level fields into `file_options` so validators/templates consume them consistently.
> 
> **Tabular/time-series** runs now always populate `file_options.number_of_columns` from the *cleaned* schema at `BaseIngestor` init (tabular-family only), and `CSVIngestor` widens default `na_values` to include `"NA"/"NULL"/"None"` for tabular-family categories (non-tabular keeps `['']`). Tests are updated/added to pin these behaviors and the template-equivalence expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 153a0618624492ec7fa6f38b7a4e3fe01d7555cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/cli/conventions.py
+++ b/tracebloc_ingestor/cli/conventions.py
@@ -99,7 +99,9 @@ DEFAULT_IMAGE_FILE_OPTIONS_BY_CATEGORY: Dict[str, Dict[str, Any]] = {
     TaskCategory.IMAGE_CLASSIFICATION:    {"target_size": [512, 512], "extension": ".jpg"},
     TaskCategory.SEMANTIC_SEGMENTATION:   {"target_size": [512, 512], "extension": ".jpg"},
     TaskCategory.OBJECT_DETECTION:        {"target_size": [448, 448], "extension": ".jpg"},
-    TaskCategory.KEYPOINT_DETECTION:      {"target_size": [448, 448], "extension": ".jpg"},
+    # keypoint_detection: no target_size default — the customer's pose model
+    # dictates input resolution, so the schema requires it top-level.
+    TaskCategory.KEYPOINT_DETECTION:      {"extension": ".jpg"},
     # No template exists yet for instance_segmentation; mirror semantic for
     # forward-compatibility, revisit when the template lands.
     TaskCategory.INSTANCE_SEGMENTATION:   {"target_size": [512, 512], "extension": ".jpg"},
@@ -257,6 +259,20 @@ def resolve(config: Dict[str, Any]) -> ResolvedConfig:
     #     other spec.file_options key behaves.
     if category == TaskCategory.TIME_TO_EVENT_PREDICTION and resolved.time_column:
         resolved.file_options.setdefault("time_column", resolved.time_column)
+
+    # 7b. Bridge top-level `target_size` (any image category — customer
+    #     override, or the only source for keypoint_detection) and
+    #     `number_of_keypoints` (keypoint_detection only — schema requires it).
+    #     Precedence is spec.file_options > top-level > category default; since
+    #     spec was already merged in step 7, top-level only fills in when spec
+    #     didn't set it.
+    if "target_size" in config and "target_size" not in spec_file_options:
+        resolved.file_options["target_size"] = list(config["target_size"])
+    if (
+        "number_of_keypoints" in config
+        and "number_of_keypoints" not in spec_file_options
+    ):
+        resolved.file_options["number_of_keypoints"] = config["number_of_keypoints"]
 
     # 8. annotation_column — keypoint_detection's existing template uses
     #    column "Annotation" (the keypoint coords carried in the CSV). The
--- a/tracebloc_ingestor/ingestors/base.py
+++ b/tracebloc_ingestor/ingestors/base.py
@@ -180,8 +180,7 @@ class BaseIngestor(ABC):
         # being sent to the backend as part of meta_data.
         if schema:
             self.file_options["schema"] = table_schema
-            if "number_of_columns" in self.file_options:
-                self.file_options["number_of_columns"] = len(table_schema)
+            self.file_options["number_of_columns"] = len(table_schema)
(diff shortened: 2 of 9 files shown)


tracebloc_ingestor/cli/conventions.py
```diff
@@ -99,7 +99,9 @@
     TaskCategory.IMAGE_CLASSIFICATION:    {"target_size": [512, 512], "extension": ".jpg"},
     TaskCategory.SEMANTIC_SEGMENTATION:   {"target_size": [512, 512], "extension": ".jpg"},
     TaskCategory.OBJECT_DETECTION:        {"target_size": [448, 448], "extension": ".jpg"},
-    TaskCategory.KEYPOINT_DETECTION:      {"target_size": [448, 448], "extension": ".jpg"},
+    # keypoint_detection: no target_size default — the customer's pose model
+    # dictates input resolution, so the schema requires it top-level.
+    TaskCategory.KEYPOINT_DETECTION:      {"extension": ".jpg"},
```
saadqbal, line 104: We have removed the target_size. is it intentional ?

tracebloc_ingestor/ingestors/base.py
```diff
@@ -180,8 +180,7 @@ def __init__(
         # being sent to the backend as part of meta_data.
         if schema:
             self.file_options["schema"] = table_schema
-            if "number_of_columns" in self.file_options:
-                self.file_options["number_of_columns"] = len(table_schema)
+            self.file_options["number_of_columns"] = len(table_schema)
```
saadqbal: Here we are always adding the number_of_columns and they are calculted from schema. is this intentional ?

saadqbal (dismissed):

tracebloc_ingestor/cli/conventions.py
divyasinghds: Intentional — keypoint_detection's target_size is dataset-specific (pose model input resolution). The old default [448, 448] didn't match what the template uses (256×256), so it was misleading rather than helpful. The schema now requires target_size top-level for category: keypoint_detection (see the new if/then block in ingest.v1.json), and the resolver bridges it into file_options. Other image categories keep their per-category defaults unchanged.

saadqbal (approved):

PR #138, opened 2026-06-01
test(e2e): end-to-end ingestion equivalence suite (all modalities, real MySQL)

LukasWodka:
## Summary
Adds the missing **end-to-end ingestion suite** — the real engine ingesting each modality's bundled `templates/` dataset into a **real MySQL**, with an in-process mock backend. Closes #134; part of epic tracebloc/backend#597.

Today every layer mocks its neighbours (CLI tests mock DB+API; `release-image` does only a schema-load smoke), so the full **validate → file-transfer → MySQL insert** path is never exercised. Running all 10 modalities through this harness showed the **engine is sound**, but **8/10 fail first-try** because the shipped configs, the `templates/` data, and the validator defaults are mutually inconsistent — exactly what breaks "copy a config, point at your data."

## What's here
- **`e2e/test_ingest_e2e.py`** — per modality: build an `ingest.yaml` matched to the bundled template, run the real engine into MySQL, assert success + rows.
- **`e2e/conftest.py`** — in-process mock backend (`CLIENT_ENV=local`) + `STORAGE_PATH` redirect; **auto-skips when no MySQL is reachable**, so the unit `pytest --cov-fail-under=95` run is unaffected (0 collected without a DB).
- **`e2e/docker-compose.yml`** + **`e2e/README.md`** — one-command local run.
- **`.github/workflows/e2e.yml`** — CI job with a MySQL service.

## Results
**7 pass** (text, time-to-event, tabular classification/regression, image, time-series, keypoint). **3 xfail** against tracking tickets (an XPASS signals the fix):

| Modality | Gap | Ticket |
|---|---|---|
| object_detection | bundled VisDrone `difficult=2` rejected by the validator | #135 |
| semantic_segmentation | mask sidecar column not wired through the declarative path | #136 |
| masked_language_modeling | template missing `tokenizer.json` | #137 |

The matched configs here differ from the shipped `examples/yaml` (which don't match the templates); aligning the examples + the validator/UX gaps is tracked in #135.

## Test plan
- Local: `docker compose -f e2e/docker-compose.yml up -d` then `pytest e2e/ -v` → **7 passed, 3 xfailed** (7s).
- Confirmed **0 tests collected** when MySQL is absent → the unit suite + coverage gate are unaffected.
- CI: the new `e2e` job runs it against a MySQL service container.

## Type of change
- [x] Test / CI (no engine behaviour change; the xfail'd gaps are fixed in #135–#137)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,55 @@
+name: E2E ingestion
+
+# Real end-to-end ingestion: runs the engine against the bundled templates into
+# a real MySQL (service container) with an in-process mock backend. Complements
+# the unit suite (tests.yml), which mocks the DB + API. See e2e/README.md.
+
+on:
+  pull_request:
+    branches: [develop, master, main]
+  push:
+    branches: [develop, master, main]
+
+permissions:
+  contents: read
+
+jobs:
+  e2e:
+    name: e2e (real MySQL)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    services:
+      mysql:
+        image: mysql:8.0
+        env:
+          MYSQL_ROOT_PASSWORD: root
+          MYSQL_DATABASE: training_test_datasets
+        ports:
+          - 3306:3306
+        options: >-
+          --health-cmd="mysqladmin ping -uroot -proot"
+          --health-interval=5s
+          --health-timeout=3s
+          --health-retries=20
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          cache: pip
+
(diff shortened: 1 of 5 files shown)


aptracebloc (approved):
Reviewed. Purely additive — new `e2e/` dir + `e2e.yml` workflow, no production code touched. `conftest.py` gates collection on MySQL reachability (`collect_ignore_glob`) so the default unit `pytest` stays green when no DB is present; in-process mock backend and the `STORAGE_PATH`→tmp redirect are sound. Tests assert `rc == 0` **and** row count > 0 — a genuine validate→file-transfer→MySQL-insert check, not a smoke test. CI green (e2e real MySQL + pytest 3.11/3.12). LGTM.

PR #146, opened 2026-06-03
test(schema): lock category enum to the engine's TaskCategory (anti-drift)

LukasWodka:
## What
Adds `test_schema_category_enum_matches_engine_categories` — asserts the `category` enum in `tracebloc_ingestor/schema/ingest.v1.json` **exactly equals** the engine's `TaskCategory.get_all_categories()`.

## Why
`cli/run.py` validates every ingest config against this schema before the engine runs, and the schema is published as the ingestion contract that downstream services validate submissions against. If the engine gains a category the enum omits, every config for it is rejected *before* ingestion starts — both in-process and at any downstream submit gate. Nothing currently ties the enum to the engine's category source-of-truth, so they can drift silently.

Found while exercising **all 10 ingest modalities end-to-end on a live install**: a few modalities were rejected at a downstream submit gate carrying a stale copy of this enum. The schema-copy sync is already guarded on the consumer side; this closes the remaining gap on the **source** side.

## Test
- ✅ Passes on `develop` (enum == TaskCategory, 11 == 11).
- ✅ Fails if a category is added to the engine but not the schema — or vice-versa (verified locally).
- Pure unit test (no MySQL / no engine run); always runs in CI.

### DIFF
--- a/tests/test_schema_validation.py
+++ b/tests/test_schema_validation.py
@@ -98,6 +98,37 @@ def test_all_task_categories_covered():
     assert not missing, f"No example for categories: {missing}"
 
 
+def test_schema_category_enum_matches_engine_categories():
+    """Anti-drift: the schema's ``category`` enum must list EXACTLY the
+    categories the engine supports (``TaskCategory``).
+
+    ``cli/run.py`` validates every ingest config against this schema before the
+    engine runs, so a category the engine supports but the enum omits would be
+    rejected before ingestion ever starts. This schema is also published as the
+    ingestion contract, so any downstream service that validates submissions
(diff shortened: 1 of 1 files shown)


saadqbal (approved):
Reviewed as assigned. `test_schema_category_enum_matches_engine_categories` is a clean bidirectional anti-drift guard — the schema's `category` enum must exactly equal `TaskCategory.get_all_categories()`, with a clear message naming drift in either direction (engine-but-not-schema / schema-but-not-engine). Test-only, green, and disjoint from the recent validator/ingestor merges (touches only test_schema_validation.py). Good regression net for the published ingestion contract. LGTM.

PR #150, opened 2026-06-06
fix(validator): stop reporting missing values as "non-numeric" (and show sample bad values)

LukasWodka:
## Summary

`DataValidator` is the schema-type check that runs for every tabular-family ingestion — tabular classification/regression, time-series forecasting, time-to-event (see `validators_mapping.py`). Its numeric checks counted **every** NaN as a "non-numeric" value:

```python
numeric_series = pd.to_numeric(series, errors="coerce")
non_numeric_count = numeric_series.isnull().sum()   # <- genuine NaN counts too
```

So a numeric (`INT`/`FLOAT`/`DOUBLE`/`DECIMAL`/`BIGINT`) column with legitimate **missing values** failed with:

> Column 'X' contains N non-numeric values

— a misleading error the user **cannot clear by inserting NaN** (NaN still counts as null). The ingestor itself stores NaN as `NULL` (`csv_ingestor._validate_csv`), so missing values are valid input; only *present-but-unparseable* values are genuinely non-numeric.

## Fix

Mirror the distinction `NumericColumnsValidator` already makes:

```python
non_numeric_mask = numeric_series.isna() & series.notna()   # present but unparseable
```

- **Missing values (NaN/empty) no longer fail** numeric validation (they're stored as NULL).
- **Genuine non-numeric values still fail** — now with up to **5 sample invalid values** in the message, so the user can see *what's* wrong (e.g. comma decimals `"1,23"`, sentinels like `"<LOD"` / `"n.d."`, stray text). Previously the error gave no examples.
- Also fixes the parallel `INT` "non-integer" check, which miscounted `NaN` via `NaN % 1 != 0`.

Example — `FLOAT` column `[1.5, NaN, "oops"]`:
- **before:** `contains 2 non-numeric values` (the NaN is miscounted; no hint what's wrong)
- **after:** `contains 1 non-numeric value(s). Sample invalid values: ['oops']`

## Tests

4 new regression tests (missing-values-valid for INT/FLOAT; missing doesn't mask a real non-numeric; sample values surfaced). Full suite green: **611 passed**.

### DIFF
--- a/tests/test_data_validator.py
+++ b/tests/test_data_validator.py
@@ -95,6 +95,41 @@ def test_float_non_numeric_fails():
 
 
 # ---------------------------------------------------------------------------
+# Missing values are NULL, not "non-numeric" (regression)
+# ---------------------------------------------------------------------------
+
+def test_int_with_missing_values_is_valid():
+    # NaN/empty in an INT column is a missing value (stored as NULL), not a
+    # non-numeric one — and not a "non-integer" one either.
+    df = pd.DataFrame({"n": [1, None, 3]})
+    assert DataValidator(schema={"n": "INT"}).validate(df).is_valid
+
+
+def test_float_with_missing_values_is_valid():
+    # Regression: a float column with genuine NaN was wrongly reported as
+    # containing "non-numeric values" — and inserting NaN could never clear it.
+    df = pd.DataFrame({"x": [1.5, None, 2.25]})
+    assert DataValidator(schema={"x": "FLOAT"}).validate(df).is_valid
+
+
+def test_missing_values_do_not_mask_real_non_numeric():
+    # Only the genuinely unparseable value is flagged; the NaN is ignored.
+    df = pd.DataFrame({"x": [1.5, None, "oops"]})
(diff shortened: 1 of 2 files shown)


saadqbal (approved):
Verified for the v0.3.3 release.

**Logic** — `_validate_int`/`_validate_float` now compute non-numeric as `numeric_series.isna() & series.notna()`, so genuine missing values (NaN/empty, stored as NULL) no longer fail; only present-but-unparseable values do. `_validate_bigint` delegates to `_validate_int`, so it's covered. The INT non-integer check is now NaN-safe (`numeric_series.notna() & (numeric_series % 1 != 0)`).

**Parity** — matches the distinction already used by `NumericColumnsValidator` (numeric_columns_validator.py:108) and `TimeToEventValidator` (time_to_event_validator.py:140); aligns DataValidator with the existing convention rather than inventing one.

**Scope** — 2 files, +32 validator / +35 tests, no scope creep. Surfacing up to 5 sample invalid values is a real UX improvement (was: no examples).

**CI** — pytest 3.11 + 3.12 and the real-MySQL e2e all green on 4692c1e; the 4 new regression tests ran inside those green runs.

LGTM.

PR #151, opened 2026-06-06
harden(validator): reject non-finite (inf) numerics + realistic messy-data scenario tests

LukasWodka:
## Summary

Proactive hardening of the tabular ingestion gate, prompted by the class of real-world messy-data bugs the `DataValidator` missing-value fix (#150) exposed. This PR closes the most clear-cut, end-to-end-proven gap: **non-finite values**.

`inf` / `-inf` survive `pd.to_numeric` (they *are* numeric), so they sailed through both numeric validators. But on the training side (`tracebloc-client`) they flow **unchanged** through `MissingValueImputationStep` (only fills NaN) and `StandardScalingStep` (only guards `std == 0`, not `inf`) straight into the model → **NaN loss, silently**. So a CSV with an `inf` cell passes ingestion and quietly poisons training.

## Changes

- `_validate_int` / `_validate_float` now reject non-finite values via a shared `_non_finite_error` helper, surfacing the offending values:
  `Column 'x' contains 2 non-finite value(s) (inf/-inf): [inf, -inf]`
- The INT "non-integer" check is now gated on finiteness (`np.isfinite`), so `inf` isn't double-reported as both non-finite and non-integer.

## Tests

Adds a realistic messy-data scenario block to `test_data_validator.py`:

| scenario | expected |
|---|---|
| `inf` / `-inf` in FLOAT and INT | **rejected** (new) |
| EU comma decimals `"1,5"` | rejected, sample value shown |
| thousands separators `"1,234"` | rejected |
| units `"95%"`, `"3kg"` | rejected, sample value shown |
| scientific notation `"1.23E+08"` | valid (mass-spec intensities) |

Full suite green: **617 passed**.

## Stacking

Stacks on **#150** (base = `fix/data-validator-missing-values-not-non-numeric`). Retarget to `develop` once #150 merges.

## Not in this PR (follow-ups from the same audit — see the findings summary)

- **Validator↔ingestor NA-token mismatch** (a file with text `NaN`/`N/A`/`null` passes `DataValidator` but `CSVIngestor._validate_csv` then raises) — needs a policy decision on aligning the two readers.
- **Column-name length > 64** (the customer contact's Problem B) — no preflight; explodes at `CREATE TABLE`.
- **Non-UTF-8 (Latin-1) files** and **`;`-delimited German CSVs** — crash / cryptic error.
- **Training-side** (`tracebloc-client`): NaN-in-label, single-class label, high-cardinality one-hot — break with no guard/test.

### DIFF
--- a/tests/test_data_validator.py
+++ b/tests/test_data_validator.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import numpy as np
 import pandas as pd
 import pytest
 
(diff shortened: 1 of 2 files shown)


saadqbal (approved):
Verified for the v0.3.4 hardening release. Builds correctly on #150 — the missing-value mask is preserved; adds `_non_finite_error` rejecting inf/-inf with sample values, and finite-gates the non-integer check so inf isn't double-counted. Covers INT/BIGINT/FLOAT/DOUBLE/DECIMAL (bigint→int, double/decimal→float delegation). Real impact: inf otherwise flows unchanged into training scaling → silent NaN loss. Green CI. LGTM.

PR #155, opened 2026-06-06
harden(ingestion): align tabular NA parsing with validator + reject over-long column names

LukasWodka:
## Summary

Two real-world ingestion **seams** found in the messy-data hardening audit (the same audit that produced #150 / #151). Both are cases where the gate lets bad data through and it bites later.

### 1. NA-token gate-lie (validator says OK → ingestor crashes)

The validators read CSVs with **plain `pd.read_csv` defaults**, so `NaN` / `N/A` / `null` / `nan` cells become `NaN`. But `CSVIngestor.read_data` used `keep_default_na=False` with a narrow `na_values=["", "NA", "NULL", "None"]`, so those same cells stayed as **text**. Result, proven in the audit:

```
DataValidator:            PASS
CSVIngestor._validate_csv: RAISES ValueError: Unable to parse string "NaN"
```

A file passes validation, then the ingestor's numeric type-conversion crashes on a token the validator treated as missing. **Fix:** tabular-family reads now use `keep_default_na=True`, so the ingestor's NA interpretation matches the validator's — what validates is what ingests. (Non-tabular categories keep the conservative empty-only behaviour.)

### 2. Column names over MySQL's 64-char limit (the customer contact's "Problem B")

CSV headers are used **verbatim** as MySQL column names (`create_table`), which are capped at 64 characters. Long proteomics/genomics headers (a semicolon-joined isoform list) exploded at `CREATE TABLE` with a raw **MySQL 1059** error. **Fix:** `create_table` now fails fast with a clear message naming the offending columns, before any DB I/O — mirroring the existing reserved-column guard.

```
3 column name(s) exceed the 64-character database column-name limit and must be
shortened: 'Protein_XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX…' (78 chars); ...
```

## Tests

- `test_csv_na_values.py`: assert `keep_default_na` per category + a behavioral test that a `NaN`/`N/A` cell in a numeric column no longer crashes `read_data` (parses as missing).
- `test_database.py`: `create_table` rejects an over-long column name with a clear error (no DB needed — runs before any I/O).

Full suite green: **613 passed**.

## Context

Independent of #150 (merged) and #151 (inf guard, open) — touches `csv_ingestor.py` + `database.py` only. Part of the broader ingestion-hardening program; remaining items (cross-cutting UTF-8/encoding handling, time-series date-locale parsing, image-pairing, training-side guards) tracked separately.

### DIFF
--- a/tests/test_csv_na_values.py
+++ b/tests/test_csv_na_values.py
@@ -48,6 +48,10 @@ def test_tabular_family_uses_wider_na_values(category, tmp_path):
 
     _, kwargs = mock_read.call_args
     assert kwargs["na_values"] == ["", "NA", "NULL", "None"]
+    # keep_default_na=True for tabular so pandas' full NA set (NaN/N/A/null/...)
+    # is recognised — matching the validators' read, which closes the gate-lie
+    # where a file passes validation but crashes the ingestor's type-conversion.
+    assert kwargs["keep_default_na"] is True
 
 
 def test_non_tabular_keeps_narrow_na_values(tmp_path):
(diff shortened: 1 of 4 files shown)


saadqbal (approved):
Verified for the v0.3.4 hardening release. `csv_ingestor` keep_default_na=True for the tabular family aligns the ingestor's NA interpretation with how the validators read (closes the gate-lie: validator PASS → ingestor crash on a 'NaN'/'N/A' token); non-tabular keeps the conservative empty-only behaviour. `database.create_table` now rejects column names over MySQL's 64-char limit before DB I/O, naming offenders, instead of a raw MySQL 1059. Files disjoint from #151/#157. Green CI. LGTM.

PR #157, opened 2026-06-06
harden(ingestion): clear error for non-UTF-8 CSVs (one pre-flight gate)

LukasWodka:
## Summary

**Tier 2 of the ingestion-hardening audit — cross-cutting encoding.**

Every CSV-reading validator opens the file with `encoding="utf-8"` inside a `try/except Exception: return None`, and `validate()` turns that `None` into **"No data found to validate"**. So a Latin-1 / Windows-1252 export — routine for German/EU lab data with umlauts (`Größe`) and accents — fails with a misleading message that points the user at the wrong problem (or, at the ingestion read, a raw `UnicodeDecodeError` traceback).

## Fix

One pre-flight UTF-8 probe in `BaseIngestor.validate_data` — the single orchestrator (`ingestors/base.py:310`) that runs **every** validator for **every** category. It decodes the CSV in 1 MB chunks before the validator loop and, on a decode error, fails fast with an actionable message:

```
'umlaut.csv' is not valid UTF-8 — a non-UTF-8 byte was found at byte 4.
Re-save the file as UTF-8 (in Excel: Save As → 'CSV UTF-8 (Comma delimited)'),
then re-ingest.
```

One chokepoint covers all categories — no per-validator edits, and **no silent latin-1 fallback** (that would mojibake umlauts and corrupt data, against the whole point of the audit). Non-CSV / missing / non-path sources are skipped and left to the validators.

## Tests (`test_ingestor_base.py`)

- Latin-1 file → rejected with a UTF-8 message
- UTF-8 file → passes
- directory / `None` / nonexistent path → skipped (no false positive)

Full suite green: **614 passed**.

## Scope

The same `encoding="utf-8"` is hardcoded in the **text/NLP dataset loaders in `tracebloc-client`** (`text_classification_dataset.py`, `masked_language_modeling_dataset.py`) — same class of bug, separate repo; tracked as a follow-up. Independent of #150/#151/#155/#156 — touches `ingestors/base.py` only.

### DIFF
--- a/tests/test_ingestor_base.py
+++ b/tests/test_ingestor_base.py
@@ -241,3 +241,29 @@ def test_context_manager_protocol():
     ing = make_ingestor()
     with ing as x:
         assert x is ing
+
+
+# ---------------------------------------------------------------------------
+# CSV encoding pre-flight (validate_data)
+# ---------------------------------------------------------------------------
+
+def test_check_csv_encoding_rejects_non_utf8(tmp_path):
+    # A Latin-1 export (German umlauts) used to surface as a misleading
+    # "No data found"; now it fails fast with a clear UTF-8 message.
+    bad = tmp_path / "umlaut.csv"
+    bad.write_bytes("Größe,label\n1,a\n".encode("latin-1"))
+    with pytest.raises(ValueError, match="UTF-8"):
+        BaseIngestor._check_csv_encoding(str(bad))
+
+
+def test_check_csv_encoding_accepts_utf8(tmp_path):
+    good = tmp_path / "ok.csv"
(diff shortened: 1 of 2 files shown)


saadqbal (approved):
Verified for the v0.3.4 hardening release. Single UTF-8 pre-flight in `validate_data` (the one orchestrator running every validator for every category); skips non-CSV/non-path/missing sources, fails fast on Latin-1/Win-1252 with an actionable 'Re-save as UTF-8' message instead of the misleading 'No data found', no silent latin-1 fallback. Green CI. LGTM. (Minor, non-blocking: the reported byte offset is chunk-relative for files >1 MB.)

PR #230, opened 2026-06-11
fix: propagate hard ingest failures in all modalities, send only inserted records

shujaatTracebloc:
## Problem

Follow-up to #223 (batch API-send failures counted as failed records, non-zero exit), which was only verified end-to-end for **token_classification**. An audit of the other 10 modalities confirmed the #223 mechanism itself covers all of them — every template routes through `BaseIngestor._flush_batch` → `APIClient.send_batch`, with no modality-specific bypass path — but found two real bugs of the same silent-success class, plus a diagnostics gap.

### 1. Five templates swallowed exceptions → exit 0 on hard failure

`image_classification`, `tabular_classification`, `tabular_regression`, `time_series_forecasting`, `time_to_event_prediction` had:

```python
except Exception as e:
    logger.error(f"{str(e)}")   # no raise
```

Any exception escaping `ingest()` — validation failure, DB error, or the fail-loud backend-registration RuntimeErrors from `base.py` (rejected edge-label / global-meta / prepare) — was logged and dropped, so the process exited **0** and the K8s Job was marked **Succeeded**. Notably, the "No data found for table name" registration failure from the original incident would still have exited 0 in these five modalities even after #223, because it surfaces as an exception, not as failed records. (#223's `sys.exit(1)` only fires when `ingest()` *returns* failed records; SystemExit bypasses the handler.) Each handler now re-raises, matching the other six templates.

### 2. Mid-batch DB failure sent the wrong records to the API

`_process_batch` paired the send with `zip(ids, batch)` — positional, truncating to `len(ids)`. `insert_batch`'s per-record fallback returns success IDs in scan order, so after a mid-batch DB failure the API send included the **DB-failed** record (a phantom backend record pointing at no MySQL row) and dropped the **last inserted** one (a committed row the platform never sees). The send list is now the batch minus DB failures, matched by `data_id` — the same rule `_flush_batch` already uses for failure accounting.

### 3. Registration-call diagnostics truncated at 100 chars

`send_global_meta_meta`, `send_generate_edge_label_meta`, `prepare_dataset`, `create_dataset` still had the pre-#223 `str(e)[:100]` pattern with no `response=` attached to the manually-raised HTTPError — a DRF 400 body was cut right after `"HTTP 400: "`. Mirrored the #223 `send_batch` fix onto all four: response attached, handler logs `HTTP <status>: <body>` capped at 2,000 chars. Failure propagation is unchanged (these already returned False / raised).

## Tests

`tests/test_modality_failure_accounting.py` (39 tests, **additive only** — no existing test touched):

- api-send-failure and clean-run accounting parameterized across **all 11 template categories** (the #223 tests ran `category=None`, which skips the file-transfer branch the 8 file-bearing categories take)
- AST structural check that every template's `except Exception` handler re-raises
- mid-batch mis-send regression guard (asserts the send carries exactly the inserted records' data_ids)
- 400-diagnostics guards for the four client methods
- guard that `create_dataset`'s raise escapes `ingest()` (it's the one registration call whose return value isn't checked)

Full suite: **955 passed, 1 xfailed**, coverage 97% (gate: 95%).

## Not included (flagged for follow-up discussion)

- `JSONIngestor` silently skips type-invalid records (CLI JSON runs exit 0 despite drops) — skip-vs-fail is a semantics decision
- templates pass `on_bad_lines: "warn"` overriding the framework default `"error"`; file-bearing categories don't run `DataValidator` on the labels CSV, so a ragged row is silently dropped

---

> [!NOTE]
> **Medium Risk**
> Changes failure propagation and batch API payload selection in the core ingest path; incorrect pairing could still cause data/platform skew, but behavior aligns with existing `_flush_batch` accounting and is heavily regression-tested.
> 
> **Overview**
> Extends the #223 “no silent success” behavior across all ingestion modalities so K8s Jobs fail when ingestion truly breaks.
> 
> **Five modality templates** (`image_classification`, tabular variants, time series, time-to-event) no longer log-and-swallow in `except Exception`; they **re-raise** after logging so validation/DB/backend registration errors exit **non-zero** (previously exit 0 / Job Succeeded).
> 
> **`_process_batch`** now builds the API payload from records that **actually inserted**, filtering out DB failures by `data_id` instead of positional `zip(ids, batch)`—fixing phantom backend rows and dropped committed rows on mid-batch DB errors.
> 
> **`APIClient`** registration helpers (`send_global_meta_meta`, edge labels, `prepare_dataset`, `create_dataset`) mirror #223’s `send_batch` fix: attach `response=` on raised `HTTPError` and log **`HTTP <status>: <body>`** (up to 2k chars) instead of `str(e)[:100]`.
> 
> Adds **`tests/test_modality_failure_accounting.py`**: per-category API failure accounting, AST check that all templates re-raise, mid-batch DB/API pairing regression, and registration error logging guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00c122e8ead7484e48a463ddaebe8cf9bed1c870. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/templates/image_classification/image_classification.py
+++ b/templates/image_classification/image_classification.py
@@ -85,7 +85,12 @@ def main():
                 logger.info("All records processed successfully")
 
     except Exception as e:
-        logger.error(f"{str(e)}")
+        logger.error(f"Ingestion failed: {str(e)}")
+        # Re-raise so the process exits non-zero — swallowing here let a
+        # hard failure (validation error, DB error, backend registration
+        # rejection raised by ingest()) end with exit code 0 and a K8s
+        # Job marked Succeeded.
+        raise
 
 
 if __name__ == "__main__":
--- a/templates/tabular_classification/tabular_classification.py
+++ b/templates/tabular_classification/tabular_classification.py
@@ -83,7 +83,12 @@ def main():
                 logger.info("All records processed successfully")
 
     except Exception as e:
-        logger.error(f"{str(e)}")
+        logger.error(f"Ingestion failed: {str(e)}")
+        # Re-raise so the process exits non-zero — swallowing here let a
+        # hard failure (validation error, DB error, backend registration
+        # rejection raised by ingest()) end with exit code 0 and a K8s
+        # Job marked Succeeded.
+        raise
 
 
 if __name__ == "__main__":
--- a/templates/tabular_regression/tabular_regression.py
+++ b/templates/tabular_regression/tabular_regression.py
@@ -83,7 +83,12 @@ def main():
                 logger.info("All records processed successfully")
 
     except Exception as e:
-        logger.error(f"{str(e)}")
+        logger.error(f"Ingestion failed: {str(e)}")
+        # Re-raise so the process exits non-zero — swallowing here let a
+        # hard failure (validation error, DB error, backend registration
+        # rejection raised by ingest()) end with exit code 0 and a K8s
+        # Job marked Succeeded.
+        raise
 
 
 if __name__ == "__main__":
--- a/templates/time_series_forecasting/time_series_forecasting.py
+++ b/templates/time_series_forecasting/time_series_forecasting.py
@@ -88,7 +88,12 @@ def main():
                 logger.info("All records processed successfully")
 
     except Exception as e:
-        logger.error(f"{str(e)}")
+        logger.error(f"Ingestion failed: {str(e)}")
+        # Re-raise so the process exits non-zero — swallowing here let a
+        # hard failure (validation error, DB error, backend registration
+        # rejection raised by ingest()) end with exit code 0 and a K8s
+        # Job marked Succeeded.
+        raise
 
 
 if __name__ == "__main__":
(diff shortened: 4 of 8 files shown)


saadqbal (approved):

templates/image_classification/image_classification.py
```diff
@@ -85,7 +85,12 @@ def main():
                 logger.info("All records processed successfully")
 
     except Exception as e:
-        logger.error(f"{str(e)}")
+        logger.error(f"Ingestion failed: {str(e)}")
+        # Re-raise so the process exits non-zero — swallowing here let a
+        # hard failure (validation error, DB error, backend registration
+        # rejection raised by ingest()) end with exit code 0 and a K8s
+        # Job marked Succeeded.
+        raise
```
aptracebloc, line 93: Can we extract common functionality here? It's evident from the explanatory comments across many files that the logic is exactly the same between templates, so we should extract and avoid duplication.

shujaatTracebloc: created a new PR for this

shujaatTracebloc: Done in #231 — the run/report/exit block now lives once in `tracebloc_ingestor.run_ingestion` and all 11 templates delegate to it.

PR #231, opened 2026-06-12
refactor: extract templates' run/report/exit block into shared run_ingestion

shujaatTracebloc:
## Why

Follow-up to the review on #230 ([comment](https://github.com/tracebloc/data-ingestors/pull/230#discussion)): the ~20-line run-and-report block (run `ingest()`, log each failed record, `sys.exit(1)` on failures, re-raise hard errors) was inlined in all 11 template scripts. The hand-maintained copies had already drifted twice:

- five templates swallowed exceptions and exited 0 on hard failures (fixed in #230)
- the failed-record log line was wrong in three different ways (MLM/text templates called `.get("filename")` on the wrapper dict and logged `None`; the tabular templates' `.get("name")` always logged `Unknown`)

## What

- **`tracebloc_ingestor/utils/template_runner.py`** — new `run_ingestion(ingestor, source, batch_size, logger)`, exported from the package root. Owns the whole contract: failed records → log each + `SystemExit(1)`; exception escaping `ingest()` → log + re-raise; clean run → success log. Also fixes the failed-record identifier: `filename` where the category has one, else `data_id` (tabular family), else `Unknown`.
- **All 11 templates** — `main()` now just builds the modality-specific ingestor and calls `run_ingestion(...)`. The per-template `try/except`, `sys.exit(1)` block, and explanatory comments are gone; `import sys` removed.
- **Structural tests migrated** (their old assertions matched the inlined source pattern this change removes; the invariant is preserved, not weakened):
  - `test_template_exits_nonzero_on_failed_records` now pins that every template imports + calls `run_ingestion` from the package root.
  - `test_template_except_handler_reraises` now pins that any reintroduced `except Exception` / bare `except` in `main()` must re-raise, and that `main()` routes through `run_ingestion`.
- **`tests/test_template_runner.py`** (new) — behavioral coverage of the helper: `SystemExit(1)` on failed records with per-record logging, re-raise of the original exception on hard errors, SystemExit passes the handler untouched, context-manager unwinding on all paths, identifier fallback chain, package-root export.

## Tests

Full suite: **962 passed, 1 xfailed**; `template_runner.py` at 100% coverage, total 97% (gate: 95%).

---

> [!NOTE]
> **Medium Risk**
> Changes how ingestion jobs fail and report errors across all templates; incorrect helper behavior would mark K8s jobs successful on partial or hard failures, though behavior is heavily tested.
> 
> **Overview**
> Extracts the duplicated run/report/exit logic from all **11** modality templates into a shared **`run_ingestion`** helper in `tracebloc_ingestor/utils/template_runner.py`, exported from the package root. Templates now only wire up their `CSVIngestor` and delegate ingestion to that helper.
> 
> The helper centralizes the K8s-oriented exit contract: log each failed record (with **`filename` → `data_id` → `Unknown`** identifiers, fixing inconsistent per-template logging), **`sys.exit(1)`** when `ingest()` returns failures, and log + re-raise hard errors instead of exiting 0. Structural template tests now assert import/call of `run_ingestion` and guard against reintroduced swallowing handlers; **`tests/test_template_runner.py`** covers the helper behavior directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f71e4a887fe4449737e6bd16cde99651ae01dc18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/templates/image_classification/image_classification.py
+++ b/templates/image_classification/image_classification.py
@@ -6,9 +6,14 @@ supporting both binary data and file-based processing.
 """
 
 import logging
-import sys
 
-from tracebloc_ingestor import Config, Database, APIClient, CSVIngestor
+from tracebloc_ingestor import (
+    Config,
+    Database,
+    APIClient,
+    CSVIngestor,
+    run_ingestion,
+)
 from tracebloc_ingestor.utils.logging import setup_logging
 from tracebloc_ingestor.utils.constants import (
     TaskCategory,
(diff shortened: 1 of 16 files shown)


aptracebloc (approved):
Nice!!

PR #354, opened 2026-07-09
master-> Fix CSV BOM handling and enhance label column resolution

saadqbal:
## Type of change
- [ ] Feature
- [ ] Bug fix
- [ ] Tech-debt / refactor
- [ ] Docs
- [ ] Security / hardening
- [ ] Breaking change

## Checklist
- [ ] Tests added / updated and passing locally
- [ ] Docs updated if behavior or config changed
- [ ] No secrets / credentials in the diff
- [ ] For security-sensitive paths: appropriate reviewer requested

---

> [!NOTE]
> **Medium Risk**
> Changes core label extraction and CSV header probing paths where mismatches previously caused silent wrong data or false validation failures; layout JSON is additive but expands the public contract surface.
> 
> **Overview**
> Fixes **silent all-NULL labels** when the manifest’s label name differs from the header only by case or whitespace (#340). A shared **`resolve_column`** rule is introduced; validators delegate through it, and **`BaseIngestor`** pins **`label_column`** to the real header key during the ingest loop (including retry on sparse JSON).
> 
> **Excel UTF-8 BOM** handling is aligned across validation and CSV ingest (#338): header probes use **`utf-8-sig`** (with **`_bom_safe_encoding`** canonicalizing UTF-8 codec aliases like `utf_8`) so BOM-prefixed first columns are not treated as missing or as distinct names that bypass duplicate-header checks.
> 
> Adds a **machine-readable per-task layout contract** (#347): **`build_layout_contract`**, committed **`schema/layout.v1.json`**, **`Sidecar`** / **`RecordFormat`** on modality specs, and tests for drift and congruence with the registry.
> 
> Package version bumps to **0.6.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c73d8c262aeb0414e13caf79f8d0aad62c87d373. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_columns.py
+++ b/tests/test_columns.py
@@ -1,0 +1,42 @@
+"""Tests for the shared column-name resolution rule (#340).
+
+``resolve_column`` is the single source of truth for matching a declared
+column name to the actual header, used by BOTH the validators
+(``BaseValidator._match_column`` delegates here) and the ingest read path
+(``BaseIngestor._resolve_label_column``). The two used to diverge — that was
+the #340 silent-null-label bug.
+"""
+
+import pandas as pd
+import pytest
+
+from tracebloc_ingestor.utils.columns import resolve_column
+from tracebloc_ingestor.validators.base import BaseValidator
+
+
+@pytest.mark.parametrize(
+    "cols,name,expected",
+    [
+        (["a", "b"], "a", "a"),                     # exact match wins
+        (["Label", "feat"], "label", "Label"),      # case-insensitive; returns original spelling
+        ([" label ", "n"], "label", " label "),     # whitespace-insensitive; returns raw spelling
+        (["Age", "Income"], "AGE", "Age"),           # fold on both sides
+        (["label"], " label ", "label"),             # whitespace on the query side too
+        (["a", "b"], "missing", None),               # nothing matches
+        ([], "x", None),                             # empty columns
+    ],
+)
+def test_resolve_column(cols, name, expected):
+    assert resolve_column(cols, name) == expected
+
+
+def test_resolve_column_accepts_pandas_columns_and_dict_keys():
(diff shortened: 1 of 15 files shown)


aptracebloc (approved):
Approving — reviewed the core-logic diff (#340 label resolution, #338 BOM, #347 layout contract), CI, and Bugbot threads. Code is correct and well-tested; the shared `resolve_column` single-source-of-truth for #340 is the right fix for the silent all-NULL-label bug. CI is effectively green: the `gate / gate` FAILURE is a **stale, superseded run** — two later `gate` runs on the same head commit (`c73d8c2`) passed.

Two **non-blocking** nits for a future touch (not blocking this release sync):

1. **`resolve_column` ambiguous-collision behavior** (`utils/columns.py`): when two headers normalize to the same key (e.g. `Label` and `label ` both present), the `{normalised: c}` comprehension silently keeps the *last* one. Duplicate-header detection lives elsewhere so real risk is low, but a defensive check (or a comment noting the last-wins behavior) would make the resolution rule fully unambiguous.

2. **`_bom_safe_encoding` cp65001 alias** (`ingestors/csv_ingestor.py`): the docstring lists `cp65001` among the UTF-8 spellings that get upgraded to `utf-8-sig`, but on some CPython versions `codecs.lookup("cp65001").name` isn't `"utf-8"`, so that specific alias may not actually be upgraded. Edge-of-edge (not a regression — the core `utf-8`/`utf_8`/`U8` cases work); worth either verifying or trimming the docstring example.

Reminder for merge: this is the develop→master **release sync** (publishes PyPI 0.6.0), so merge as a **merge commit** (not squash) per RELEASING.md.

PR #378, opened 2026-07-20
feat(backfill): metadata builder for pre-cutover datasets (di#360)

divyasinghds:
## What & why

Datasets ingested **before** the federated-alignment cutover (di#360 / backend#1037) carry no per-column `feature_stats` and no enriched schema on the global-metadata channel, so the backend's combine-time alignment checks can't reconcile them. This adds a **metadata-only backfill builder** so an operator can rebuild that metadata per client, per dataset table, for pre-cutover datasets — **without re-ingesting rows**.

`tracebloc_ingestor/metadata_backfill.build_dataset_metadata(database, table_name, *, category, label_column=None, data_format=None, file_options=None)` returns the same `{schema, meta_data}` payload a fresh ingest ships:

- **Numeric `feature_stats`** — computed from SQL aggregates (`COUNT` / `SUM` / `SUM(x*x)` / `MIN` / `MAX`) over the persisted table, not a chunked re-read.
- **Categorical vocab** — a bounded `GROUP BY` (drops columns past the cardinality cap; per-value min-count suppression applied by the ingestor).
- **Enriched schema** — the framework `label` column marked `role: "target"` for supervised tasks.

It **reuses** `CSVIngestor._schema_payload` / `_collect_run_metadata` rather than reimplementing the shapes, so a backfilled dataset's metadata is byte-identical to a freshly-ingested one (guarded by the integration tests).

## Scope — deliberately no API send

Per the agreed scope, this only **builds** the payload. Wiring it to the backend metadata endpoint (and iterating the client's dataset tables) is a separate step, left as a hook — `build_dataset_metadata` returns the dict for a caller to send.

## What is / isn't recomputed

| Fact | Source |
|---|---|
| enriched schema, numeric `feature_stats`, categorical vocab | **recomputed from the table** (the block-causing alignment facts) |
| target handling (regression target kept as `label`; class label excluded) | derived from `category` + `label_column` |
| time-series `timezone` / `sampling_frequency` | **not** recomputed (needs the ordered timestamp series) — WARN-only |
| uploader-declared scalars (`color_mode`, `language`, `time_unit`, …) | pass-through via `file_options` when the caller has them; else omitted — WARN-only |

## Correctness details

- `count` is the non-null count; all aggregates ignore nulls (`WHERE col IS NOT NULL`), matching the ingest path's `dropna()`. All-null columns are omitted.
- Framework/bookkeeping columns (`id`, `created_at`, `data_id`, …) never contribute stats/vocab. The `label` column is the exception: its numeric stats are kept for regression-class tasks (emitted under `label`, matching the live target re-key) and skipped otherwise.
- `SUM` of an INT column returns `Decimal`; scalars are coerced to native JSON numbers, keeping the column's real type (INT → integer min/max, like the live ingest's `.item()`).

## Tests

`tests/test_metadata_backfill.py`:
- **unit** (fake connection, hand-built table): framework/label exclusion, regression target kept as `label`, classification label excluded, int-vs-float + `Decimal` coercion, all-null omission, categorical cardinality cap.
- **integration** (in-memory SQLite): `build_dataset_metadata` end-to-end for regression (target stats under `label`, `id` excluded, categorical vocab) and classification (class label excluded from stats, `role: "target"` still set).

## Notes

- **Stacked on `feat/360-feature-stats`** (base of this PR) — it reuses that branch's enriched-schema / feature_stats shapes. Merge after #360.
- Storage-agnostic: it produces the payload; where the backend stores it (the ongoing Mongo → SQL move for `GlobalMetaData`) doesn't affect this builder.

---

> [!NOTE]
> **Medium Risk**
> Touches federated alignment metadata shapes and SQL over production dataset tables; correctness depends on parity with live ingest, but changes are additive with no automatic API send.
> 
> **Overview**
> Adds **`metadata_backfill.build_dataset_metadata`** so operators can rebuild the **`{schema, meta_data}`** global-metadata payload for **already-ingested** dataset tables **without re-reading or re-ingesting CSV rows**—addressing pre–federated-alignment datasets that lack **`feature_stats`** and enriched schema for backend combine checks.
> 
> The builder runs **SQL aggregates** (`COUNT` / `SUM` / `SUM(x²)` / `MIN` / `MAX`) for numeric sufficient statistics and a **bounded `GROUP BY`** for categorical vocab, with the same exclusion rules as live ingest (framework columns, classification **`label`** omitted from stats, regression target stats under **`label`**, cardinality cap, null handling). It **injects** those accumulators into **`CSVIngestor`** and calls **`_schema_payload`** / **`_collect_run_metadata`** so output shape matches a fresh ingest; **API upload is out of scope** (caller sends the returned dict). v1 does **not** recompute time-series **`timezone`** / **`sampling_frequency`**; uploader scalars can be passed via **`file_options`**.
> 
> **`csv_ingestor`** gains a comment documenting that **`_feature_stats_acc`** / **`_categorical_acc`** names and shapes must stay stable for backfill. **`tests/test_metadata_backfill.py`** adds unit tests for the SQL helpers and SQLite integration tests for regression vs classification end-to-end output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 288bd23ab42e345fc09f0c0d475aaae79dc2d80a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_metadata_backfill.py
+++ b/tests/test_metadata_backfill.py
@@ -1,0 +1,227 @@
+"""Tests for metadata_backfill — recomputing a dataset's {schema, meta_data}
+payload from an already-ingested table via SQL, without re-ingesting rows.
+
+Two tiers:
+* unit — the SQL-shaping helpers (_numeric_feature_stats / _categorical_counts)
+  against a hand-built SQLAlchemy Table + a fake connection, so the exclusion /
+  coercion / cardinality logic is covered with no database;
+* integration — build_dataset_metadata end-to-end over a real in-memory SQLite
+  table, asserting it reuses the ingestor's shapes (enriched schema with
+  role:"target", numeric feature_stats, categorical vocab).
+"""
+
+from __future__ import annotations
+
+from collections import Counter
+from decimal import Decimal
+
+import pytest
+from sqlalchemy import (
+    Column,
+    Float,
+    Integer,
+    MetaData,
+    String,
+    Table,
+    create_engine,
+    text,
+)
+
+from tracebloc_ingestor import metadata_backfill as mb
+from tracebloc_ingestor.config import Config
+
+
+# ---- fakes ---------------------------------------------------------------
+
+
+class _FakeResult:
+    def __init__(self, one=None, all_=None):
+        self._one = one
+        self._all = all_
+
+    def one(self):
+        return self._one
+
+    def all(self):
+        return self._all
+
+
+class _FakeConn:
+    """Returns canned aggregate results in execute() call order."""
+
+    def __init__(self, results):
+        self._results = list(results)
+
+    def execute(self, _stmt):
+        return self._results.pop(0)
+
+
+def _table(cols):
+    md = MetaData()
+    type_map = {"int": Integer, "float": Float, "str": String}
+    return Table("t", md, *(Column(n, type_map[t]) for n, t in cols.items()))
+
+
+# ---- unit: _numeric_feature_stats ----------------------------------------
+
+
+def test_numeric_stats_excludes_framework_and_nonnumeric_keeps_regression_label():
+    # schema order drives the query order: id(framework→skip), label(regression
+    # target→keep), age(int feature), feat(float feature), city(string→skip).
+    schema = {
+        "id": "BIGINT",
+        "label": "FLOAT",
+        "age": "INT",
+        "feat": "FLOAT",
+        "city": "VARCHAR(255)",
+    }
+    table = _table({"id": "int", "label": "float", "age": "int", "feat": "float"})
+    conn = _FakeConn(
+        [
+            _FakeResult(one=(3, Decimal("60"), Decimal("1400"), 10.0, 30.0)),  # label
+            _FakeResult(one=(3, Decimal("60"), Decimal("1400"), 18, 42)),  # age (INT)
+            _FakeResult(one=(3, 6.0, 14.0, 1.0, 3.0)),  # feat
+        ]
+    )
+    stats = mb._numeric_feature_stats(conn, table, schema, "tabular_regression")
+
+    assert set(stats) == {"label", "age", "feat"}  # id + city excluded
+    # sum/sum_sq always float; INT column keeps integer min/max, float stays float.
+    assert stats["label"] == {
+        "count": 3,
+        "sum": 60.0,
+        "sum_sq": 1400.0,
+        "min": 10.0,
+        "max": 30.0,
+    }
+    assert stats["age"]["min"] == 18 and isinstance(stats["age"]["min"], int)
+    assert stats["feat"]["min"] == 1.0 and isinstance(stats["feat"]["min"], float)
+    assert isinstance(stats["age"]["sum"], float)  # Decimal coerced
+
+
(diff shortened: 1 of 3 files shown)


LukasWodka (commented):
Really nice work on this — injecting the SQL-computed accumulators into a real `CSVIngestor` and letting its own `_schema_payload` / `_collect_run_metadata` shape the output (instead of re-deriving the payload) is exactly the right call, and it's what makes the byte-identical guarantee credible. The docstrings and the "what is / isn't recomputed" table are lovely. 👌

**One heads-up worth sorting before line-level review:** the branch looks like it predates a few things that have since landed on `develop` (it's ~7 commits behind), so the diff is showing a lot of already-merged work as new, and a straight merge would roll a couple of things back:

- `data_id_strategy` default flips back `content_hash → uuid` — that's the #350 default-flip, so we'd lose the retried-Job dedup behavior.
- The whole `feature_stats` / `_schema_payload` / `categorical_vocab` machinery shows up as new but already merged via #361 (`feat/360-feature-stats`), so it'll conflict.
- It also drops `utils/csv_dialect.py` and the #371/#376 validator changes now on develop.

Could you rebase onto the latest `develop` and force-push? Once you do, the diff should collapse to basically just `metadata_backfill.py` + `test_metadata_backfill.py` (the real new work), which makes this much easier to review. Everything below is against that new module.

**A couple of small correctness things on `metadata_backfill.py`:**

- **`SUM(x*x)` vs the live path's float64 square.** The live accumulator squares in float64 (`(fvals**2).sum()`), but the SQL here does `func.sum(column * column)`. For INT columns MySQL squares in integer/DECIMAL, which can (a) drift from the float64 result — so `sum_sq` may not stay byte-identical — and (b) risk a BIGINT overflow on large values. Casting the column to float inside the aggregate would keep it in lockstep with the ingest path.
- **The categorical `GROUP BY` cap is applied after loading all rows.** `_categorical_counts` does `.all()` then drops if `len(rows) > cap`, so a high-cardinality free-text/id column pulls every distinct value into memory first. A `.limit(cap + 1)` on the query bounds that.

**Tests:** coverage is great (the int/float/`Decimal` and regression-vs-classification cases especially). One gap — the integration/parity tests run on in-memory SQLite, but the correctness claims (SUM→`Decimal`, integer squaring, type coercion) are really MySQL behaviors SQLite won't exercise. A parity test that ingests a fixture the normal way and asserts `build_dataset_metadata` reproduces the same payload **on MySQL** (gated to the real-MySQL job) would lock the byte-identical promise down where it matters.

**Tiny stuff:**
- `_MAX_CATEGORICAL_CARDINALITY` is duplicated from `CSVIngestor._MAX_CATEGORICAL_CARDINALITY` — maybe reference the original so they can't drift.
- Since backfill leans on `CSVIngestor`'s private `_feature_stats_acc` / `_categorical_acc` / `_schema_payload`, a one-line note on those in `CSVIngestor` ("metadata_backfill depends on these") would protect it from a future rename.

None of this is major once the rebase is done — the core is solid. Happy to pair on the rebase if the #361 overlap gets hairy. Thanks Divya! 🙏

LukasWodka (approved):

PR #380, opened 2026-07-21
fix(objdet): resolve annotation by stem so extension-bearing filename works (#379)

LukasWodka:
## Summary

`object_detection` ingestion silently **skipped every record** (Job exits 9, 0 ingested) whenever the manifest `filename` column carried the image extension (e.g. `a.jpg`) — the exact format the CLI parity fixture `objdet-ok/labels.csv` ships and the template README documents as supported. The dry-run and **all** cluster validators passed first, so the failure only surfaced mid-Job as `Source annotation not found: …/annotations/a.jpg`.

Root cause: `_find_src("annotations", filename, ".xml")` kept the recognised `.jpg` extension instead of swapping to `.xml`, so it looked for `annotations/a.jpg` (never present; the annotation is `a.xml`). The **File Pairing validator pairs image↔XML by on-disk stem** (`Path.stem`), so validation passed while transfer failed — a validation/transfer inconsistency, and a violation of the README's documented *"with or without extension"* contract.

Fix: add a `force_extension` mode to `_find_src` that strips a recognised trailing extension off the manifest value and appends the sidecar's fixed extension, used at both annotation-resolution sites (the `object_detection` factory and `annotation_transfer`). Both `a.jpg` and the bare stem `a` now resolve to `annotations/a.xml`, matching `Path.stem` and the documented contract. The default (non-force) branch is byte-for-byte identical, so image / text / mask / keypoint / semseg lookups are unchanged.

Scope note: only the object_detection **annotation** sidecar was affected. `semantic_segmentation` masks already strip via `_find_mask_src`; image lookups correctly keep the manifest extension. No docs or fixture change needed — the code was the only thing wrong (docs and the `objdet-ok` fixture were already correct).

## Related

Closes #379. Same class as the image_classification `filename`-vs-`image_id` gap (separate chip), where the CLI local walk green-lights a dataset the cluster then rejects for a file-resolution reason it could surface up front.

## Type of change
- [x] Bug fix

## Test plan

- Reproduced the bug: seeded `images/a.jpg` + `annotations/a.xml`, drove `map_file_transfer(OBJECT_DETECTION, {"filename": "a.jpg"}, {"extension": ".jpg"})` → pre-fix skipped with `Source annotation not found: …/annotations/a.jpg`; post-fix stages both `a.jpg` and `a.xml`. The bare-stem workaround (`filename="a"`) still works.
- New regression tests (all confirmed to **fail on pre-fix code, pass post-fix**):
  - `test_map_object_detection_extension_bearing_filename` — end-to-end objdet with `x.jpg`, asserts both files staged **and** the stored record is identical to the bare-stem form (`filename="x"`, `extension=".jpg"` — no `.jpg` leaks into the DB row).
  - `test_map_object_detection_extension_bearing_filename_jpeg` — same for a non-`.jpg` recognised extension (`.jpeg`).
  - `test_annotation_transfer_resolves_stem_from_extension_bearing_filename` — standalone `annotation_transfer` path.
  - `test_find_src_force_extension_{swaps_real_extension,bare_stem,keeps_non_extension_dot}` — the `_find_src` stem/bare/internal-dot (`image.001`) edges.
- Full suite: `1784 passed, 1 xfailed` (the pre-existing leading-zeros INT xfail). `black --check` clean.
- Adversarially verified the #239 path-traversal guard is preserved in the `force_extension` branch (`_safe_join` still runs on the transformed value; absolute/`..` inputs still raise).

## Deployment notes

Ships in the ingestor image. Existing installs pick up the fix on the next ingestion run once `images.ingestor.tag` is moved to the new digest (see `RELEASING.md`) — no chart change required.

## Checklist
- [x] Tests added / updated and passing locally
- [x] Docs updated if behavior or config changed — N/A (fix makes code match the already-correct docs)
- [x] No secrets / credentials in the diff
- [ ] For security-sensitive paths: appropriate reviewer requested — N/A (path-traversal guard verified preserved)

---

> [!NOTE]
> **Medium Risk**
> Changes core file-resolution for object-detection annotation staging; scope is narrow (annotation sidecar only) with extensive tests and no change to default image lookups.
> 
> **Overview**
> Fixes **object detection** ingestion when manifest `filename` includes the image extension (e.g. `x.jpg`): annotation lookup now resolves **`annotations/<stem>.xml`** instead of incorrectly looking for `annotations/x.jpg`, which caused every record to be skipped after validators passed.
> 
> Adds **`force_extension`** to `_find_src`: for sidecars it strips only a **recognised** image extension from the manifest value, then appends the sidecar extension (`.xml`). Values like `image.001` stay intact so annotations remain `image.001.xml`. Wired into **`object_detection`** transfer pre-check and **`annotation_transfer`** when the source path is unresolved; default `_find_src` behaviour for images/text/masks is unchanged.
> 
> Regression tests cover `_find_src` edge cases, standalone `annotation_transfer`, and end-to-end `map_file_transfer` for `.jpg` / `.jpeg` with extension-bearing filenames.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f363b2d12c58a73edf705b01df6850b5766ac927. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/file_transfer.py
+++ b/tracebloc_ingestor/file_transfer.py
@@ -163,10 +163,16 @@ def _find_src(
     """
     cfg = cfg or config
     if force_extension:
-        # Mirror _has_extension's rsplit so the stem matches Path.stem exactly
-        # (what FilePairingValidator pairs on). A value with no recognised
-        # extension — a bare stem, or an internal dot like ``image.001`` — is
-        # kept whole so we never strip a non-extension segment.
+        # Strip only a RECOGNISED trailing extension (the same _has_extension
+        # gate the image lookup uses). For the jpg/jpeg/png images objdet ships
+        # (FileTypeValidator enforces that set) this equals the image file's
+        # Path.stem — what File Pairing / Pascal VOC XML pair on — so validation
+        # and transfer resolve the same annotation. This is deliberately NOT
+        # Path.stem: a value with no recognised extension — a bare stem, or an
+        # internal dot like ``image.001`` — is kept whole, so we never strip a
+        # non-extension segment (Path.stem would wrongly turn ``image.001`` into
+        # ``image``). The two only diverge for a non-recognised final suffix
+        # (e.g. ``a.tar.gz``), which objdet images can't have.
         stem = filename.rsplit(".", 1)[0] if _has_extension(filename) else filename
         filename_with_ext = f"{stem}{extension}"
     else:
--- a/tracebloc_ingestor/file_transfer.py
+++ b/tracebloc_ingestor/file_transfer.py
@@ -122,7 +122,12 @@ def _has_extension(filename: str) -> bool:
 
 
 def _find_src(
-    subdirectory: str, filename: str, extension: str, cfg: Optional[Config] = None
+    subdirectory: str,
+    filename: str,
+    extension: str,
+    cfg: Optional[Config] = None,
+    *,
+    force_extension: bool = False,
 ):
     """Resolve a source file in `SRC_PATH/<subdirectory>/`.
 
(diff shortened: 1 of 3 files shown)


tracebloc_ingestor/file_transfer.py
```diff
@@ -136,11 +141,38 @@ def _find_src(
...
+      validator (which pairs by on-disk stem) had passed. The manifest format
+      is documented as working "with or without extension"; this keeps the
+      annotation side of the transfer honouring that contract.
     """
     cfg = cfg or config
-    filename_with_ext = (
-        filename if _has_extension(filename) else f"{filename}{extension}"
-    )
+    if force_extension:
+        # Mirror _has_extension's rsplit so the stem matches Path.stem exactly
```
saadqbal: Small accuracy nit on the comment: this doesn't match `Path.stem` *exactly* — `_has_extension` only strips a **recognised** trailing extension (jpg/jpeg/png/xml/txt/text), whereas `Path.stem` (what FilePairing/PascalVOCXML pair on) strips any final suffix. So `a.tar.gz` → force keeps `a.tar.gz`, but `Path.stem` → `a.tar`. Not reachable for objdet today (images ship jpg/jpeg/png, enforced by FileTypeValidator), so no bug — just worth softening the "exactly" claim so a future maintainer doesn't lean on it for a non-recognised image ext.

LukasWodka: Good catch — fixed in f363b2d (comment-only). You're right it isn't `Path.stem` in general: `_has_extension` strips only a *recognised* trailing extension, so `a.tar.gz` is kept whole where `Path.stem` → `a.tar`. Reworded to scope the equivalence to the jpg/jpeg/png set objdet ships (enforced by FileTypeValidator) and to note the gate is *deliberately* not `Path.stem` — it keeps `image.001` whole, which `Path.stem` would wrongly truncate to `image`. No logic change.

saadqbal (approved):

PR #381, opened 2026-07-21
fix(ingest): reclaim staged source after verified load (~2x PVC)

LukasWodka:
## What

The ingestor **copies** each file-bearing dataset's staged sidecars from `SRC_PATH/<subdir>/` into the final table dir `DEST_PATH` (`_copy_file_with_retry`) but never removed the staging copy afterward. So every image / detection / segmentation ingest left **two** copies of the data on the shared PVC — ~2× disk per ingest (#346).

- New `file_transfer.reclaim_source(cfg)` — `rmtree`s the `SRC_PATH` staging tree.
- `BaseIngestor._ingest_with_lock` calls it right before returning, **only** when `dataset_registered is True` **and** `failed_records` is empty.

This closes the leak at the source and replaces the CLI's extra `rm -rf` cleanup pod (tracebloc/cli#167), and also covers the helm-driven ingest path (which never went through the CLI reclaim).

## Why it's safe

- **Reclaim only after a fully verified, clean load.** `dataset_registered` flips `True` only after DB rows are committed **and** the backend summary registration succeeds. On any failure/partial run the source is deliberately kept for retry/inspection — mirroring the existing compensating-delete branch, whose comment already says staged files stay and are "overwritten on re-run". No per-file `shutil.move` (that would delete source mid-load, before verification — unsafe).
- **Runs under the table lock** (released in `ingest()`'s `finally`), so it can't race a concurrent ingest of the same table.
- **Heavily guarded** — `reclaim_source` refuses to delete a directory that *is* or *contains* the freshly-written table, and never touches the shared root:
  - unset/empty `SRC_PATH` (tabular etc.) → no-op;
  - `SRC_PATH` not an existing dir → no-op;
  - `SRC_PATH == STORAGE_PATH` (shared-PVC root) → skip;
  - `SRC_PATH == DEST_PATH`, or `DEST_PATH` **inside** `SRC_PATH` (the helm-direct "staged straight into the root" layout) → skip — strictly no worse than today;
  - `SRC_PATH` inside `DEST_PATH` → skip.
- **Best-effort:** a filesystem error while removing the tree is logged and swallowed — a leftover staging copy must never turn a green ingest red.

## Files

- `tracebloc_ingestor/file_transfer.py` — `reclaim_source` helper (+ `YELLOW` import).
- `tracebloc_ingestor/ingestors/base.py` — gated call before `return failed_records`.
- `tests/test_file_transfer_reclaim.py` — unit tests for the helper (happy path + every guard + best-effort).
- `tests/test_ingest_source_reclaim.py` — gate tests: reclaim fires on clean success, and is skipped when a record fails (dataset still registers) or the summary send fails.

## Test plan

- `pytest tests/test_file_transfer_reclaim.py tests/test_ingest_source_reclaim.py` → **12 passed**; `reclaim_source` is fully covered (no missing lines).
- Full unit suite: **1785 passed, 1 xfailed**, plus the 4 new gate/helper tests. The only failures locally are 4 pre-existing `float/int overflow` message-assertion tests, which fail identically on `origin/develop` without this change (a Python 3.9-local vs CI 3.11+ pandas/numpy artifact — CI runs 3.11/3.12).
- `black --check` clean on all changed/added files. (`base.py` shows two unrelated pre-existing formatting hunks from a newer local black; left untouched to avoid churn — CI does not run black.)

Closes #346

---

> [!NOTE]
> **Medium Risk**
> Deletes directories on the shared PVC, but only via a strict opt-in path gate; residual risk is CLI staging-layout drift or a guard bug deleting the wrong tree.
> 
> **Overview**
> Fixes **~2× PVC usage** (#346) by removing the CLI staging tree after ingest copies sidecars into the table directory.
> 
> Adds **`file_transfer.reclaim_source`**, which deletes only when resolved `SRC_PATH` exactly matches `STORAGE_PATH/.tracebloc-staging/<TABLE_NAME>`. It uses **`realpath` + `commonpath`** guards (symlinks, sibling user dirs, other tables, overlap with `DEST_PATH`), **warns** on staging-layout drift under `.tracebloc-staging`, and is **best-effort** so failures never fail a successful ingest.
> 
> **`BaseIngestor._ingest_with_lock`** invokes reclaim after registration when **`dataset_registered` and `failed_records` is empty** (still under the table lock).
> 
> New tests cover reclaim guards/observability and the ingest gate (success vs partial vs summary failure).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69b407b9bfcaab5ed71159c9b68dc47c5f43ffbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/file_transfer.py
+++ b/tracebloc_ingestor/file_transfer.py
@@ -440,10 +440,19 @@ def map_file_transfer(
             record.pop(key, None)
 
 
-# The isolated, tracebloc-created staging subtree the CLI stage pod writes to
-# (SharedRoot/.tracebloc-staging/<table>; tracebloc/cli teardown.py). It is the
-# ONLY location reclaim_source will delete — a hidden, tracebloc-owned dir that
-# is provably a throwaway copy, never a customer's own data.
+# The isolated, tracebloc-created staging subtree the CLI stage pod writes to:
+# SharedRoot/.tracebloc-staging/<table>/ (tracebloc/cli internal/push/teardown.go
+# + data_ingest_cluster.go). reclaim_source deletes ONLY that per-table dir — a
+# hidden, tracebloc-owned throwaway copy, never a customer's own data.
+#
+# CROSS-REPO CONTRACT: this name — and the per-table layout the gate below
+# expects (SRC_PATH == SharedRoot/.tracebloc-staging/<TABLE_NAME>, which the CLI
+# realises by staging images: under .../<table>/images/) — is shared with the
+# CLI and nothing here pins it. If the CLI's staging layout changes, this must
+# change in lockstep or reclaim silently no-ops and the #346 ~2x-disk leak
+# returns on a green ingest. Mitigation until a shared constant / cross-repo
+# integration check exists (PR #381 review): the gate WARNs when a source lands
+# under this tree but not at the expected per-table path, so drift is loud.
 STAGING_DIRNAME = ".tracebloc-staging"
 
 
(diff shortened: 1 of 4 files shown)


tracebloc_ingestor/file_transfer.py
```diff
@@ -437,3 +438,161 @@ def map_file_transfer(
...
+    # Binding to cfg.TABLE_NAME (not merely "somewhere under .tracebloc-staging")
+    # stops a stray SRC_PATH — another dataset's staging, or a <table> symlink
+    # that realpaths into the tree — from rmtree'ing a sibling's data on the
+    # shared PVC (Bugbot #381). `expected` is the literal per-table path (NOT
+    # realpath'd), so a `<table>` that is itself a symlink elsewhere fails the
+    # equality against the fully-resolved src and is left alone. A blank table
+    # name, or one that escapes the staging subtree, skips.
+    staging_root = os.path.normpath(os.path.join(storage_real, STAGING_DIRNAME))
+    table = (cfg.TABLE_NAME or "").strip()
+    expected = os.path.normpath(os.path.join(staging_root, table))
```
saadqbal, line 576: This gate is the whole ballgame, and it silently couples us to how the CLI stages. Reclaim only fires when `SRC_PATH` resolves to *exactly* `STORAGE_PATH/.tracebloc-staging/<TABLE_NAME>` — but in the CLI path `SRC_PATH` is `dirname(images:)` (`_resolve_config`, run.py:288), so this only matches if the CLI stages images to `…/.tracebloc-staging/<table>/images/`. Nothing in this repo pins that, and `_resolve_config`'s own docstring example (`/data/images` → parent `/data`) wouldn't match. If the CLI convention ever drifts, reclaim just no-ops at info-level and #346 stays open while the ingest is green — the leak comes back invisibly. Worth a shared constant / integration check across the two repos, or at least a louder log when a file-bearing clean load produced no reclaim.

LukasWodka: Good catch on the silent coupling — addressed in 69b407b.

**Louder log (the invisible-leak concern).** The skip is now split by intent:
- A source that resolves **under** `.tracebloc-staging` but **not** to this table's expected dir — a CLI staging-layout drift (e.g. images staged under `…/<table>/data/images/` → `SRC_PATH=…/<table>/data`), or a cross-table `SRC_PATH` — now logs **WARNING**: *"…is under {staging_root} but is not this table's expected dir {expected} — NOT reclaiming, so the ~2x-disk staging copy (#346) may persist even though the ingest succeeded. If this is the CLI-staged path, its layout has drifted…"*
- A source **outside** the tree (helm user dir, storage root, symlink resolving away) stays **INFO** — a deliberate skip we don't want to false-alarm on every helm ingest.

Pinned by tests: `test_warns_loudly_on_staging_layout_drift` and `test_user_dir_skip_stays_quiet`.

Re your `_resolve_config` example (`/data/images` → `/data`): agreed it wouldn't match — and it *shouldn't* reclaim `/data`; that's the outside-the-tree INFO path. The dangerous case is a drift that stays **within** `.tracebloc-staging`, and that now shouts.

**Cross-repo contract.** Documented the coupling at `STAGING_DIRNAME` (references the CLI side — `internal/push/teardown.go` + `data_ingest_cluster.go` — and notes it must change in lockstep). The shared-constant / cross-repo integration check you suggested is the right durable fix, but it spans both repos; I've flagged it in-code as a follow-up and can open a tracking issue for it.

saadqbal (approved):

PR #392, opened 2026-07-22
fix(validators): stop doubling "Validator" in log lines (D1)

LukasWodka:
Validation logs read `Data Validator Validator successfully passed` — `validator.name` already ends in "Validator", so the literal was appended twice. Fixed in all three spots (passed / failed / error). Log-string only; 93 validator tests green. Backlog **D1**.

---

> [!NOTE]
> **Low Risk**
> Log and error-string wording only; no validation or ingestion logic changes.
> 
> **Overview**
> Fixes duplicated **Validator** in preflight user-facing strings: `validator.name` is already values like `Data Validator`, so appending another **Validator** produced `Data Validator Validator successfully passed` and `Data Validator Validator failed`.
> 
> **`validate_data` in `base.py`** now uses `{validator.name} successfully passed`, `{validator.name} failed`, and `{validator.name} error` without the extra word; a short comment documents why.
> 
> A **regression test** docstring in `test_data_validator.py` is updated to expect `Data Validator failed` instead of the doubled form. The rest of the diff is incidental formatting (line wraps, blank lines) with no behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f53f7ac57cd91d7995ff1e4611ec372470c47614. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_data_validator.py
+++ b/tests/test_data_validator.py
@@ -8,10 +8,10 @@ import pytest
 
 from tracebloc_ingestor.validators.data_validator import DataValidator
 
-
 # ---------------------------------------------------------------------------
 # validate() top-level flow
 # ---------------------------------------------------------------------------
+
 
 def test_no_schema_passes_with_warning():
     result = DataValidator().validate(pd.DataFrame({"a": [1]}))
@@ -73,7 +73,7 @@ def test_loads_from_json(tmp_path):
 
     Surfaced by an end-to-end cluster ingestion: a 20-record JSON file with
     explicit schema {id INT, age INT, score FLOAT, active BOOL, label
-    VARCHAR(20)} failed with "Data Validator Validator failed: No data found
+    VARCHAR(20)} failed with "Data Validator failed: No data found
     to validate" before any record was read.
     """
     p = tmp_path / "d.json"
@@ -186,9 +186,7 @@ def test_loads_from_json_scalar_array_fails_not_silently_passes(tmp_path):
     """
     p = tmp_path / "scalars.json"
     p.write_text("[1, 2, 3]")
-    result = DataValidator(schema={"id": "INT", "label": "VARCHAR(8)"}).validate(
-        str(p)
-    )
+    result = DataValidator(schema={"id": "INT", "label": "VARCHAR(8)"}).validate(str(p))
     assert not result.is_valid
     assert "No data found" in result.errors[0]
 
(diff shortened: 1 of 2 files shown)


shujaatTracebloc (approved):

shujaatTracebloc (commented):
## Review — stop doubling "Validator" in log lines (D1)

Good catch on the doubled word, and the three touch-points (passed / failed / error) are all covered. But the fix over-corrects — it's right for the validators whose name ends in "Validator" and introduces the *opposite* wording bug for the ones whose name doesn't.

### Change request (blocking) — 10 of 26 validators lose the word entirely
`validator.name` is an `__init__` default per validator, and they're inconsistent. **16** end in "Validator" (e.g. `"Data Validator"`, `"Pascal VOC XML Validator"`, `"Image Resolution Validator"`) — for these the old string doubled it and this PR fixes it. But **10 do not**:

```
BIO Label · Contrastive Pairs · Ingestable Records · Keypoint Annotation ·
Keypoint Visibility · Label Column · Mask Id Column · Sentence Pair ·
Tab-Separated Record · Text Content
```

For these, the old `"{name} Validator ..."` read correctly (`"Ingestable Records Validator successfully passed"`); after this change they become:

- `"Ingestable Records successfully passed"`
- `"Text Content failed: ..."`
- `"Mask Id Column error: ..."`

i.e. the user-facing preflight output now drops "Validator" for ~40% of validators — trading the doubling bug for a dropped-word bug. (These are the same user-facing strings the CLI preflight surfaces, e.g. the validator list in the #379 report — `Ingestable Records` is one of them.)

**Two ways to fix cleanly:**

1. **Conditional append (minimal, future-proof)** — don't touch 26 name strings; normalize at the log site so it's correct regardless of how any current/future validator is named:
   ```python
   label = validator.name if validator.name.endswith("Validator") else f"{validator.name} Validator"
   # then: f"{GREEN}{label} successfully passed{RESET}", f"{BOLD}{label} failed: ...", f"{label} error: ..."
   ```
2. **Normalize the names** — make all 26 `.name` defaults consistent (drop the "Validator" suffix everywhere, keep the log appending it, or vice-versa). Cleaner long-term but a 26-file diff and easy to regress when someone adds a validator.

I'd go with **(1)** — one small helper line covers all three strings and can't drift as validators are added.

### Verified / non-issues
- The de-doubling itself is correct for the 16 "…Validator"-named validators, and the `error:` string dropping the leading `"Validator "` reads fine (`"Data Validator error: …"`).
- The `pytest (3.11)` red on this PR was a **flaky test** (`test_image_validator_min_size.py::test_too_small_result_still_reports_corrupt_files`, auto-detection order/Pillow-version sensitive) — unrelated to this diff; a re-run went green. Worth its own de-flake ticket, not a blocker here.

### Nit (non-blocking)
`tests/test_data_validator.py` carries a lot of incidental `black` reformatting (blank lines, line wraps, arg re-wrapping) beyond the one docstring string this PR needs to change. Not wrong if the repo enforces black, but it mixes formatting churn into a one-line-fix PR and enlarges the review surface.

_🤖 Reviewed with [Claude Code](https://claude.com/claude-code)_

PR #416, opened 2026-07-30
ci(release): let the release train own the version tag

LukasWodka:
Companion to tracebloc/release-train#7. Fixes the tag-push failure described in tracebloc/backend#1345 (option 1, Lukas's call; Asad, this touches your #402 chain so flagging you as assignee).

## The bug this fixes

`auto-release-on-master.yml` creates the release tag as `github-actions[bot]`:

```
git tag -a "${TAG}" "${SHA}" -m "${TAG}"
git push origin "${TAG}"
```

but this repo's tag ruleset (*Protect v\* release tags — supply-chain trust root*) restricts `creation` on `refs/tags/v*` to `OrganizationAdmin`, `release-managers` and `release-python`. The bot is in none of them, so that push is rejected and the job dies under `set -euo pipefail`.

It has never been seen because the workflow's only run so far (2026-07-29) took the "nothing to release" branch — `v0.7.8` already existed — so the push path never executed. `__version__` is now **0.7.9 on develop** vs **0.7.8 on master**, so the next real release would have been the first to try it, and fail.

## What changes

**1. `auto-release-on-master.yml` is removed.** Both of its jobs move elsewhere:
- *Tagging* → the release train, which now carries `version_file: tracebloc_ingestor/__init__.py` and cuts `vX.Y.Z` on the prod hop as Lukas, who bypasses the ruleset. Same mechanism already proven on `cli`, `design-system` and `client`.
- *Triggering the image* → the tag push itself. `release-image.yml` already listens on `push: tags: ['v*.*.*']`, and a PAT-pushed tag **does** fire workflows — the explicit `gh workflow run` dispatch only existed because a `GITHUB_TOKEN`-pushed tag doesn't.

**2. `release-image.yml` gains a `verify-published` gate.** This preserves the invariant #402 was built around — *an image only ships for a build that published cleanly* — which the old design got implicitly from chaining on a successful `publish-master`. The train pushes the tag right after the promotion merge while `publish-master` is still running, so that guarantee now has to be explicit: the new job polls PyPI for the exact version (up to 10 min, since the tag can land before the upload and the CDN lags), and **fails closed** if it never appears. `release` gets `needs: verify-published`.

Net effect: the guarantee is stronger than before — it asserts the artifact is genuinely on PyPI rather than trusting that an upstream workflow concluded successfully.

## Verification

- YAML parses; both jobs present; every `run:` block passes `bash -n`.
- Version matcher tested against the real file on all three branches → `0.7.9` / `0.7.8` / `0.7.8`, and a negative test confirms it ignores dependency pins like `requests==2.31.0`.
- Gate probe tested live both ways: `tracebloc-ingestor==0.7.8` resolves (gate passes), `9.9.9` 404s (gate holds, then fails closed).

## Sequencing note

`workflow_run` workflows execute from the **default branch**, so `auto-release-on-master.yml` keeps running from `master` until this reaches it via the normal develop → staging → master promotion. In that window a master push may log one failed run of it. Harmless — the train's tag does the real work, and that workflow was already going to fail for exactly this reason. It goes quiet once the deletion lands on master.

---

> [!NOTE]
> **Medium Risk**
> Changes release CI sequencing and supply-chain gates; mis-timed tags or a broken PyPI poll could block images until operators re-drive, but runtime ingestor behavior is unchanged.
> 
> **Overview**
> **Release tagging moves out of this repo** because `auto-release-on-master.yml` would push `v*` tags as `github-actions[bot]`, which the protected-tag ruleset blocks. That workflow is **removed**; the org **release train** cuts `vX.Y.Z` on the prod hop and the tag push alone triggers `release-image.yml`.
> 
> **`release-image.yml` now enforces “PyPI before image” explicitly.** A new `verify-published` job (no elevated permissions) requires a `vX.Y.Z` ref, polls PyPI for up to ~10 minutes, and **fails closed** if the package never appears—covering the race where the train tags before `publish-master` finishes. The `release` job depends on that gate.
> 
> **Concurrency is keyed on the tag** (`inputs.ref || github.ref_name`) so a manual `workflow_dispatch` cannot race a tag-triggered run and double-push/sign or fight over `gh release create`.
> 
> **Docs** (`RELEASING.md`, `.cursor/BUGBOT.md`) describe train-owned tags, the PyPI gate, manual re-drive, and new troubleshooting rows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77f97ec8824d85276668c6039b77fd3aee8e9895. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LukasWodka:
bugbot run

.github/workflows/release-image.yml
```diff
@@ -44,7 +44,49 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
saadqbal, line 52: `group: release-image-${{ github.ref }}` one line up doesn't cover the dispatch path — on `workflow_dispatch` `github.ref` is `refs/heads/master`, not `inputs.ref`. You already handle exactly this for `tags:`, `checkout`, `REF` and `TAG`, and the metadata-action comment at L124-131 spells the gotcha out; this line just got left behind.

It matters more after this PR: the gate turns the pre-build window from seconds into up to 10 minutes of silence, and RELEASING.md:218 tells operators to `gh workflow run release-image.yml -f ref=v${VERSION}` when a run looks stuck. Two runs then race — different groups, and `cancel-in-progress` is false since there's no `pull_request` trigger. Both push `:X.Y.Z`/`:X.Y`/`:X` from independent non-reproducible builds (`provenance: mode=max`, `sbom: true`), and because the train pushes a bare tag with no Release both take the `gh release create` branch at L296-298 — so one 422s *after* it already pushed and cosign-signed an image.

`group: release-image-${{ inputs.ref || github.ref_name }}`.

.github/workflows/release-image.yml
```diff
@@ -44,7 +44,49 @@ concurrency:
...
 jobs:
+  # Only cut an image for a version that actually published. The old chain got
+  # this implicitly: auto-release-on-master.yml ran on a *successful*
+  # publish-master (sdist install + schema probe) and only then tagged and
+  # dispatched. The release tag now comes from the release train
+  # (tracebloc/backend#1345), which pushes it right after the promotion merge
+  # while publish-master is still running — so the guarantee has to be stated
+  # explicitly here instead of inherited from the trigger.
+  verify-published:
+    timeout-minutes: 15
```
saadqbal, line 63: Worth `permissions: {}` on this job. It's one unauthenticated curl to a public URL, but with no job-level block it inherits `contents: write` + `packages: write` + `id-token: write` + `attestations: write` from L32-36 — and an OIDC token minted here carries the `release-image.yml` workflow_ref, i.e. the exact identity the cosign command in the file header tells consumers to trust. Splitting the gate into its own job is what makes this free; may as well take it.

.github/workflows/release-image.yml
```diff
@@ -44,7 +44,49 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.check.outputs.version }}
```
saadqbal: Nothing reads `needs.verify-published.outputs.version` — `release` re-derives the ref at six sites (L97, L134-136, L161, L257). Either drop the `outputs:` block, or wire it through so the version the gate cleared is provably the one the image is built from. If you wire it, note this one is `v`-stripped and every consumer wants the tag form.

.github/workflows/release-image.yml
```diff
@@ -44,7 +44,49 @@ concurrency:
...
+            *) echo "::error::ref '$REF' is not a vX.Y.Z tag" >&2; exit 1 ;;
+          esac
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+          # publish-master.yml uploads to PyPI on the master push; the tag can
+          # land first, and PyPI's CDN lags a little behind an upload. Poll
+          # rather than race, and FAIL CLOSED: no published sdist means no
+          # image, which is the whole point of this gate.
+          for attempt in $(seq 1 30); do
+            if curl -fsS -o /dev/null "https://pypi.org/pypi/tracebloc-ingestor/${version}/json"; then
```
saadqbal: `curl -fsS` has no timeout of any kind — there's no default `--max-time`, and `-f`/`-s`/`-S` don't add one. I checked against a listener that accepts the connection and never replies: it just hangs indefinitely. So a post-handshake stall blocks *before* the `sleep 20`, the loop never reaches attempt 2, and `timeout-minutes: 15` kills the step — which means the carefully-worded `::error::` at L85 never prints and you get GitHub's generic "exceeded maximum execution time" instead. Fail-closed still holds, but the one signal this gate exists to give is exactly the one you lose.

`--connect-timeout 10 --max-time 30`, and maybe `--retry 2` so a transient 502 on attempt 30 isn't indistinguishable from a real 404.

.github/workflows/release-image.yml
```diff
@@ -44,7 +44,49 @@ concurrency:
...
+              echo "tracebloc-ingestor ${version} is on PyPI - proceeding to build the image."
+              exit 0
+            fi
+            echo "  attempt ${attempt}/30: ${version} not on PyPI yet, waiting 20s"
+            sleep 20
+          done
+          echo "::error::tracebloc-ingestor ${version} never appeared on PyPI (10 min). Refusing to publish an image for a version that did not publish - check the 'Publish Master Package' run for this commit." >&2
```
saadqbal: On the framing in the PR body — this proves the version *string* is on PyPI, not that the tagged commit is the one that published it. The old chain bound image ⇄ green `publish-master` ⇄ `head_sha` three ways; here there's no commit binding at all, so a tag sitting on a commit whose tree differs from what uploaded still clears the gate. On the train path you're fine (`$merge_sha` is what publish-master builds) — it's the manual paths that lose it, and RELEASING.md itself notes v0.7.6–v0.7.8 got tagged off develop. A `git merge-base --is-ancestor` after checkout would close it if you think it earns the line. Not blocking — every reachable path is still better than today, where a hand-cut tag builds an image with no PyPI check whatsoever.

Separate but related: nothing re-drives a stranded tag now. `push: tags` can't re-fire for an existing ref and the ruleset blocks delete-and-re-push, so a run that dies here needs a human to notice — the deleted `tag exists && no Release → re-dispatch` block is gone. The manual dispatch is documented and works; it's the noticing that isn't automated.

saadqbal (commented):
Design's right — moving the tag to the train is the correct fix, and stating the #402 invariant explicitly instead of inheriting it from the trigger is the right call. Three things I'd fix before merge (inline) plus the runbook.

**RELEASING.md isn't touched and is now wrong.** §5 (L102-104) still names `auto-release-on-master.yml` as the thing that reads `__version__` and cuts the tag, and describes idempotency that no longer exists anywhere. L15 says the image and the PyPI package release **independently** — `needs: verify-published` makes that false in one direction now. `.cursor/BUGBOT.md:25` repeats the same claim, which matters more than it looks since that's what steers bugbot on this repo. Worth a troubleshooting row for a gate timeout too, since that's a brand-new symptom.

Two nits not worth their own threads: the `case` glob at L68 isn't a semver check (`*` matches dots and letters, so `1x.2y.3junk` passes the thing whose error says "is not a vX.Y.Z tag"), and `${REF#v}` makes the `v` optional so `-f ref=0.7.9` clears the gate and then dies in checkout instead; and `sleep 20` is unguarded on the 30th iteration, so you burn 20s after the verdict is already decided.

One thing I checked that turned out to be a non-issue, so you don't have to: PyPI's JSON API normalises the version in the path — `.../5.0-rc.1/json` returns 200 — so an rc tag wouldn't hang the gate for 10 minutes.

Not approving yet on the strength of the three inline ones. All cheap.

saadqbal (approved):

PR #437, opened 2026-08-03
feat(#203-A3): ingestor writes file-bearing assets under ds_<hex>

LukasWodka:
## #203 Track A3 — RFC-0003 D9 phase 2 / D16

Today's [#255](https://github.com/tracebloc/client-runtime/pull/255) (scoped mount) and [#569](https://github.com/tracebloc/tracebloc-engine/pull/569) (engine path) both re-keyed dataset access onto the physical `ds_<hex>` handle. But the ingestor still **hard-refused** file-bearing categories under `PER_INGESTION_TABLES`, because their assets landed in the shared **label** tree (`STORAGE_PATH/<label>`) where a same-label re-ingest would overwrite an earlier dataset's files. This PR moves the file tree onto the same handle and lifts the refusal.

### Change
- **`config.py`** — `DEST_PATH` now keys on a new `DEST_TABLE` (defaults to the `TABLE_NAME` label, injectable via `set_dest_table`). This is the *one knob*: every file-copy primitive, the text profiler, the duplicate validator, and the source-reclaim path already read `DEST_PATH` off the run config, so they all follow automatically.
- **`ingestors/base.py`** — when `per_ingestion_tables`, inject the `ds_<hex>` handle: `config.set_dest_table(physical_table_name)`. File-bearing assets now land in `STORAGE_PATH/ds_<hex>/`, matching the scoped mount (#255) and `get_dataset_path` (#569). The file-bearing refusal is removed.
- **tests** — flipped the "refuses file-bearing" test to assert the handle injection; added a real-`Config` `DEST_PATH` test. **Flag off is byte-for-byte unchanged** (`DEST_TABLE` = label, `set_dest_table` never called).

### Why it matters
Unlocks image / object-detection / semantic-segmentation / keypoint datasets (the bulk of tracebloc's use-cases) under per-ingestion isolation, and completes the D9-phase-2 file half that #255 + #569 were built to enable. It's the ingestor change the on-HOLD [client#490](https://github.com/tracebloc/client/pull/490) v0.8.0 image is waiting on.

### Scope note
The file **table lock** intentionally stays keyed on the label — that's pre-existing #408 behavior (the row store already moved to `ds_<hex>` while the lock stayed on the label); it's a concurrency guard, not a data-isolation unit, so it's out of scope here.

### Verified
Full ingestor suite **1903 passed, 1 xfailed**; `black` clean.

Targets `develop`.

---

> [!NOTE]
> **Medium Risk**
> Changes on-disk layout for file-bearing datasets when PER_INGESTION_TABLES is on; misalignment with scoped mounts or backend path resolution could strand or mis-associate assets, though flag-off behavior is unchanged.
> 
> **Overview**
> With **PER_INGESTION_TABLES** enabled, file-bearing ingests no longer error at construction; instead the run repoints the on-disk file tree from the shared label directory to the same **`ds_<hex>`** handle used for the MySQL row store.
> 
> **Config** introduces **`DEST_TABLE`** (defaults to the user label) and **`set_dest_table`**, so **`DEST_PATH`** becomes `STORAGE_PATH/<DEST_TABLE>`. **BaseIngestor** calls **`set_dest_table(physical_table_name)`** when the flag is on, which pulls file copy, text profiling, duplicate checks, and source reclaim along without separate changes. **`TABLE_NAME`** and flag-off behavior stay as today.
> 
> Tests now assert handle injection for image-classification under the flag and **`DEST_PATH`** repointing; package version **0.8.0 → 0.8.1**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 110d3d8f67bdf94fc5bb529778a4d129af79d1ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_per_ingestion_tables.py
+++ b/tests/test_per_ingestion_tables.py
@@ -10,6 +10,7 @@ ingest summary carries the physical handle so the backend can persist it
 """
 
 import json
+import os
 import uuid
 from unittest.mock import MagicMock, patch
 
@@ -23,7 +24,6 @@ from tracebloc_ingestor.ingestors import base as base_mod
 from tracebloc_ingestor.ingestors.base import BaseIngestor
 from tracebloc_ingestor.ingestors.csv_ingestor import CSVIngestor
 
-
 # ── Config flag ──────────────────────────────────────────────────────────────
 
 
@@ -34,8 +34,16 @@ def test_flag_defaults_off(monkeypatch):
 
 @pytest.mark.parametrize(
     "raw,expected",
-    [("1", True), ("true", True), ("YES", True), ("on", True),
-     ("0", False), ("", False), ("no", False), ("off", False)],
+    [
+        ("1", True),
+        ("true", True),
+        ("YES", True),
+        ("on", True),
+        ("0", False),
+        ("", False),
+        ("no", False),
+        ("off", False),
+    ],
 )
 def test_flag_env_parsing(monkeypatch, raw, expected):
     monkeypatch.setenv("PER_INGESTION_TABLES", raw)
@@ -44,9 +52,14 @@ def test_flag_env_parsing(monkeypatch, raw, expected):
 
 def test_flag_override_beats_env(monkeypatch):
     monkeypatch.setenv("PER_INGESTION_TABLES", "1")
-    assert Config(EDGE_ENV="local", PER_INGESTION_TABLES=False).PER_INGESTION_TABLES is False
+    assert (
+        Config(EDGE_ENV="local", PER_INGESTION_TABLES=False).PER_INGESTION_TABLES
+        is False
+    )
     monkeypatch.delenv("PER_INGESTION_TABLES", raising=False)
-    assert Config(EDGE_ENV="local", PER_INGESTION_TABLES=True).PER_INGESTION_TABLES is True
+    assert (
+        Config(EDGE_ENV="local", PER_INGESTION_TABLES=True).PER_INGESTION_TABLES is True
+    )
 
 
 # ── Ingestor naming ─────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


saqlainsyed007 (approved):
Verdict: APPROVE.

Reviewed the full-file context, not just the hunks. The change is correct and consistent with the row-store side of RFC-0003 D9 phase 2.

What I verified:
- **DEST_PATH repoint** — `DEST_TABLE` defaults to `TABLE_NAME`, so flag-off is byte-for-byte unchanged; `set_dest_table(physical_table_name)` is the only writer, and every consumer (`file_transfer` copy primitives, `text_profile`, `duplicate_validator`, `reclaim_source`) reads `DEST_PATH` off the same run config, so the single knob genuinely propagates.
- **Path safety** — `physical_table_name` is always `ds_<uuid4().hex>` (hex-only), so no traversal/injection surface is introduced by moving the file tree.
- **Reclaim** — `_reclaim_source` still keys the source gate on `TABLE_NAME` (the CLI's `.tracebloc-staging/<label>` dir) and only uses `DEST_PATH` for the overlap backstop; staging (`<label>`) and dest (`ds_<hex>`) don't overlap, so reclaim still fires correctly.
- **Table lock** — stays on the label via `TableLock(self.table_name, …)`, matching the scope note; concurrency-only, not a data-isolation unit.
- **Refusal removal** — clean: `_FILE_BEARING_CATEGORIES` is still used (base.py:547, 1031), no dead import.
- **Version bump** — 0.8.0→0.8.1 is single-sourced in `__init__.py` and parsed by `setup.py:_read_version`; nothing else to update.
- **Tests** — real assertions (handle injection via `set_dest_table.assert_called_once_with`, and a real-`Config` DEST_PATH repoint asserting label untouched), not vacuous.

Non-blocking observation (no change requested): `set_dest_table` permanently mutates the shared `Config` override with no reset path. This is safe because `cli/run.py` builds one Config → one ingestor per process; if a future caller ever reused one Config across a flag-on then flag-off ingestor, the stale `ds_<hex>` DEST_TABLE would leak into the flag-off run. Worth a one-line comment noting the single-run-per-config invariant, but not reachable today.

PR #440, opened 2026-08-03
fix(#203-A3): reclaim ds_<hex> file tree on failed per-ingestion ingest (data-ingestors#439)

LukasWodka:
## Fixes the #439 PVC-leak finding on #437

**Finding (Bugbot Medium):** a failed file-bearing ingest under `PER_INGESTION_TABLES` leaves assets under its unique `ds_<hex>` handle. Unlike the flag-off label tree (a same-label re-run overwrites it), the retry mints a **fresh** handle → the orphaned files are never overwritten and nothing reclaims them → lasting PVC leak for image-sized payloads.

> Verified real. Dormant in prod *only* because `PER_INGESTION_TABLES` defaults off — so this is a **blocker to close before the enablement flip**, not a live prod bug. (Note: #437 *lifted* the refusal, so file-bearing now writes; the earlier "refused cleanly" read was pre-A3.)

## Fix
- **`file_transfer.reclaim_dest_tree()`** — guarded, best-effort `rmtree` of `DEST_PATH`. **Safety (opt-in, provable):** deletes only when the basename is a `ds_<hex>` handle (never a user label) **and** it resolves to a *direct child* of `STORAGE_PATH`; `realpath` before check + delete; every error logged + swallowed so a cleanup failure can never mask the ingest's original error. Mirrors `reclaim_source`'s posture.
- **`base.py` failure handler** calls it for per-ingestion file-bearing runs — **not** gated on `inserted_records` (files can land before any row). Flag-off is a safe no-op (the handle guard only matches `ds_<hex>`).
- **+4 tests**: removes the handle tree; leaves the flag-off label tree; refuses a non-direct-child handle; no-op on missing dest.

## Scope
Closes the **graceful-failure** path (the common case). **Hard kills** (OOMKilled/SIGKILL) bypass the except-branch — those orphaned file trees are the **edge-side husk sweep**'s job (client-runtime, exactly like the DB husk tables it already reaps); flagged for a separate client-runtime change, not done here.

Full suite **1907 passed**; `black` clean. `__version__` 0.8.1 → 0.8.2. Targets `develop`.

---

> [!NOTE]
> **Medium Risk**
> Deletes files on the shared PVC during ingest failure handling; mitigated by strict `ds_<hex>` and direct-child guards plus the `dataset_registered` check, but a guard bug could cause data loss on registered datasets.
> 
> **Overview**
> Fixes **PVC leaks** when a file-bearing ingest fails under `PER_INGESTION_TABLES`: each run writes assets under a unique `ds_<hex>` tree that retries never overwrite, so orphaned files could accumulate until this change.
> 
> Adds **`reclaim_dest_tree()`** in `file_transfer` — a best-effort, guarded `rmtree` of `DEST_PATH` that only runs when the basename matches `ds_<hex>` and the path is a **direct child** of `STORAGE_PATH` (with `realpath` checks). Flag-off label directories and nested paths are left alone; errors are swallowed so cleanup cannot mask the original failure.
> 
> Wires it into **`BaseIngestor`’s compensating-failure path** for per-ingestion, file-bearing runs that did **not** register (`not dataset_registered`), mirroring the existing row delete guard so a late post-registration failure does not delete live dataset files.
> 
> Adds unit and ingest integration tests for the happy reclaim path, safety no-ops, and the registration guard. Bumps package version **0.8.1 → 0.8.2**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b51fa136e624817a30f56aeea113d703e8153b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_file_transfer_reclaim.py
+++ b/tests/test_file_transfer_reclaim.py
@@ -351,3 +351,63 @@ def test_user_dir_skip_stays_quiet(tmp_path, monkeypatch):
     cfg = _cfg(userdir, storage, table="cats_dogs_train")
     assert file_transfer.reclaim_source(cfg) is False
     assert calls["info"] == 1 and calls["warning"] == 0  # deliberate, quiet
+
+
+# ---------------------------------------------------------------------------
+# reclaim_dest_tree — a FAILED per-ingestion run's dest tree (#437 / #439)
+# ---------------------------------------------------------------------------
+# Flag-on file trees key on a UNIQUE ds_<hex> handle, so a failed run's assets
+# are never overwritten by a retry (fresh handle) and nothing else reclaims
+# them. reclaim_dest_tree removes them on the failure path — but ONLY when it
+# can prove the dir is a ds_<hex> handle directly under STORAGE_PATH.
+
+_HANDLE = "ds_" + "a" * 32
+
+
+def _dest_cfg(storage, dest_table):
+    """Config whose DEST_PATH = storage/<dest_table> (set_dest_table override)."""
+    cfg = Config(TABLE_NAME="my_label")
+    cfg.STORAGE_PATH = str(storage)
+    cfg.set_dest_table(dest_table)
+    return cfg
+
+
+def test_reclaim_dest_removes_per_ingestion_handle_tree(tmp_path):
+    storage = tmp_path / "shared"
+    dest = storage / _HANDLE
+    dest.mkdir(parents=True)
+    (dest / "img.jpg").write_bytes(b"x")
+    assert file_transfer.reclaim_dest_tree(_dest_cfg(storage, _HANDLE)) is True
+    assert not dest.exists()
+
+
+def test_reclaim_dest_leaves_flag_off_label_tree(tmp_path):
+    # flag off: DEST_PATH = storage/<label>; the basename is not a ds_<hex>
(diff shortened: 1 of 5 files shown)


saqlainsyed007 (approved):
Approved. Solid, well-reasoned, and well-tested.

The `reclaim_dest_tree` guard envelope is provably safe — deletes only a `^ds_[0-9a-f]{32}$` basename that is a *direct child* of `STORAGE_PATH`, realpath'd before both the check and the delete, best-effort so cleanup can never mask the ingest's original error. The `not dataset_registered` gate (correctly *not* gated on `inserted_records`, since files can land before any row) mirrors the row compensating-delete and is covered by a genuine mutation test.

One non-blocking note: the two-generals residual window the row-delete comment documents applies to files too — if the backend registered but the client-side `send_ingest_summary` still raised, `dataset_registered` stays False and the `ds_<hex>` tree the backend points at gets reclaimed. It's symmetric with an already-accepted risk (narrow: 409-idempotent resend, POST retried), but for files a re-run mints a fresh handle so the registered dataset isn't repopulated. Worth one sentence in the `reclaim_dest_tree` docstring acknowledging the window persists for files — not a code change.

Optional: the unrelated Black reformatting in test_ingestor_base.py could live in a separate sweep to keep this diff focused.

PR #443, opened 2026-08-04
fix(ci): make the prod PyPI publish explicit + least-privilege (backend#1427)

LukasWodka:
## Summary

`publish-main.yml` (renamed from `publish-master.yml` by #444) runs `twine upload dist/*` with **no `--repository-url`**, so every push to the prod branch publishes the `tracebloc-ingestor` sdist + wheel to **production PyPI** (`upload.pypi.org`) while the step was named *"Publish to GitHub Packages"* — hiding what it did. It also had no `permissions:` block and used `checkout@v3`, on a **public** repo. #444 fixed the branch name (`master` → `main`) but not the publish. Flagged as defect 3 in the backend#1427 release-path audit.

`tracebloc-ingestor` is a live, actively-published PyPI package (37 releases; `0.8.2` shipped via this workflow on 2026-08-04), so this hardens the path rather than removing it.

## Change

- Make the PyPI target **explicit** (`--repository-url https://upload.pypi.org/legacy/`) and rename the step for what it does.
- Add a least-privilege `permissions: contents: read` block.
- Bump `actions/checkout@v3` → `@v4`.

## Why no approval gate (revised after Bugbot)

My first version bound this to a `production-pypi` environment with required reviewers. **Bugbot correctly flagged that this breaks the image build**: `release-image.yml` (tag-triggered) has a `verify-published` job that polls PyPI for the new version for ~10 minutes and **fails closed**, and its image `build` job `needs` it. A manual approval that held the upload past 10 minutes would fail the image build on every release. And the push to the prod branch is *already* a train-gated, reviewed promotion, so a second human approval is redundant as well as harmful. The environment gate is removed.

## ⚠️ Follow-up for Lukas — the real fix for the long-lived token

The remaining risk — a long-lived `PYPI_API_TOKEN` in a public repo's repo secrets, readable by any workflow — is best closed by **PyPI Trusted Publishing (OIDC)**: it removes the token entirely and ties publishing to this specific workflow, with **no blocking step** (so it doesn't hit the image-build timing problem). That needs a PyPI-side trusted-publisher config + switching to `pypa/gh-action-pypi-publish` with `id-token: write`. Worth a dedicated follow-up.

## Coordination

Rebased onto develop after **#444** (backend#1428 Change 2) renamed `publish-master.yml` → `publish-main.yml`. Complementary to **#1422** (systemic publish-trigger lockdown) — the `on: push` trigger is left untouched.

## Test plan

- `actionlint` clean (pre-existing `setup-python@v4` notice left as-is, consistent with the sibling `publish-dev.yml`).

Refs #1427 (parent epic tracebloc/backend#1405).

---

> [!NOTE]
> **Medium Risk**
> Touches the live production PyPI publish workflow and secret usage on a public repo; behavior is clarified and hardened rather than redirected, with follow-up noted for OIDC trusted publishing.
> 
> **Overview**
> Hardens **production package publishing** in GitHub Actions after audit finding that `publish-main.yml` ran `twine upload` with no `--repository-url` (defaulting to **production PyPI**) while the step was labeled *Publish to GitHub Packages*.
> 
> The prod workflow now **names the step accurately**, passes an explicit `https://upload.pypi.org/legacy/` URL, and adds **`--skip-existing`** so re-runs after partial uploads or unchanged versions don’t fail with “file already exists.” Both `publish-main.yml` and `publish-dev.yml` get **`permissions: contents: read`**, **`cancel-in-progress: false`** (with rationale that in-flight publishes must not be cancelled), and bumps to **`actions/checkout@v4`** / **`setup-python@v5`**. Inline comments document why a manual approval gate was **not** added (would break `release-image.yml`’s ~10 minute PyPI poll). `release-image.yml` comments/errors now refer to **`publish-main`** and the renamed workflow title.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e5d29c337f1b873d82cea765897f0932426c8c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/publish-main.yml
+++ b/.github/workflows/publish-main.yml
@@ -7,17 +7,33 @@ on:
 
 concurrency:
   group: publish-main-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  # push to the prod branch is the only trigger, so there is never a PR run to
+  # cancel; a publish in flight must not be cancelled anyway. State it plainly.
+  cancel-in-progress: false
+
+# Least privilege: this job publishes to PyPI with PYPI_API_TOKEN, not with the
+# workflow token, so it needs nothing beyond reading the checked-out source.
+permissions:
+  contents: read
 
 jobs:
   publish:
     timeout-minutes: 20
     runs-on: ubuntu-latest
+    # NOT gated behind a required-reviewer environment, deliberately. release-image.yml
+    # (tag-triggered) has a verify-published job that polls PyPI for this version for
+    # ~10 minutes and FAILS CLOSED, and its image build `needs` it -- so a manual
+    # approval that held this upload past 10 minutes would break the image build on
+    # every release (Bugbot, backend#1427). The push to the prod branch is already a
+    # train-gated, reviewed promotion, so a second human approval here is redundant as
+    # well as harmful. The remaining risk -- a long-lived PyPI token in a public repo's
+    # repo secrets -- is best closed by PyPI Trusted Publishing (OIDC), which removes
+    # the token entirely with no blocking step; flagged for follow-up, not a settings pass.
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.11'
 
@@ -41,10 +57,18 @@ jobs:
         run: |
           python -c "from tracebloc_ingestor.cli.run import _load_schema; print(_load_schema()['title'])"
 
-      - name: Publish to GitHub Packages
+      - name: Publish to production PyPI
+        # This step was named "Publish to GitHub Packages", but `twine upload`
+        # with no --repository-url targets upload.pypi.org -- so every push to
+        # the prod branch published to PRODUCTION PyPI while claiming otherwise
+        # (backend#1427). The target is now explicit and the step is named for
+        # what it does. --skip-existing makes the step idempotent: a re-run
+        # after a partial upload (wheel landed, sdist hit a transient error), or a
+        # push that did not bump the version, becomes a green no-op instead of a
+        # permanent HTTP 400 "File already exists" with no clean re-run path.
         run: |
           python -m pip install twine
-          twine upload dist/*
+          twine upload --skip-existing --repository-url https://upload.pypi.org/legacy/ dist/*
         env:
           TWINE_USERNAME: '__token__'
           TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -8,17 +8,22 @@ on:
 
 concurrency:
   group: publish-dev-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  cancel-in-progress: false
+
+# Parity with publish-main.yml: this workflow builds + smoke-tests only (it does
+# not upload), but least privilege still applies on a public repo.
+permissions:
+  contents: read
 
 jobs:
   publish:
     timeout-minutes: 20
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.11'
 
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-main.yml
```diff
@@ -41,10 +55,15 @@ jobs:
...
+      - name: Publish to production PyPI
+        # This step was named "Publish to GitHub Packages", but `twine upload`
+        # with no --repository-url targets upload.pypi.org -- so every push to
+        # the prod branch published to PRODUCTION PyPI while claiming otherwise
+        # (backend#1427). The target is now explicit and the step is named for
+        # what it does.
         run: |
           python -m pip install twine
-          twine upload dist/*
+          twine upload --repository-url https://upload.pypi.org/legacy/ dist/*
```
saqlainsyed007: `twine upload` here has no `--skip-existing`. Since this workflow runs on every push to `main`, a duplicate-version or partial upload permanently reds the job: if the wheel uploads and the sdist then hits a transient network error, any re-run finds the wheel already on PyPI and twine exits non-zero with HTTP 400 "File already exists", with no clean re-run path without manual PyPI intervention. Same failure if any commit lands on `main` without bumping the version. Since this PR is rewriting this exact step to harden the path, `twine upload --skip-existing --repository-url https://upload.pypi.org/legacy/ dist/*` closes the trap. This is the one item I'd consider a blocker.

.github/workflows/publish-main.yml
```diff
@@ -9,12 +9,26 @@ concurrency:
...
+    # train-gated, reviewed promotion, so a second human approval here is redundant as
+    # well as harmful. The remaining risk -- a long-lived PyPI token in a public repo's
+    # repo secrets -- is best closed by PyPI Trusted Publishing (OIDC), which removes
+    # the token entirely with no blocking step; flagged for follow-up, not a settings pass.
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python
         uses: actions/setup-python@v4
```
saqlainsyed007: The PR bumps `actions/checkout` v3 → v4 for currency but leaves `actions/setup-python@v4` (current major is v5), so the modernization is only partial — one action pinned current, an adjacent one two majors behind. Minor; I see the PR notes this was left to match `publish-dev.yml`, so feel free to defer, but bumping both keeps the file internally consistent.

.github/workflows/publish-main.yml
```diff
@@ -9,12 +9,26 @@ concurrency:
   group: publish-main-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
saqlainsyed007: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}` can never be true here — this workflow's only trigger is `push` to `main`, so `github.event_name` is always `push`. It's copy-paste dead code from a PR-triggered template and reads as if concurrent publishes get cancelled under some condition when they never do. A literal `false` (or dropping the key) states the actual behavior. Pre-existing, not introduced by this PR.

saqlainsyed007 (commented):
Reviewed at high effort. The core change is correct — explicit `--repository-url`, least-privilege `permissions`, honest step name, and the inline rationale all check out, with no correctness regression introduced by the diff. A few hardening/consistency notes; only the `--skip-existing` one is worth treating as a potential blocker. Inline comments below.

Two findings land in files outside this diff, so I couldn't attach them inline:

- **`release-image.yml` — stale operator guidance.** The `verify-published` failure message (around line 117) still says `check the 'Publish Master Package' run` and the comment at line 94 references `publish-master.yml`. After #444's rename and this PR's `name:` change to "Publish Package to PyPI (prod)", an operator following that message during a release-blocking failure looks for a run/workflow that no longer exists. Worth a follow-up touch since it's on the same publish path.
- **`publish-dev.yml` — not hardened.** The `permissions: contents: read` block and `checkout@v4` bump were applied only here; the sibling dev publish workflow (also `on: push`, same public repo) still has no `permissions` block and uses `checkout@v3`, so the two now diverge in security posture.

saqlainsyed007 (approved):
Approving — all five findings from the previous review are addressed in 0e5d29c, and I checked the new changes for regressions (none):

- **`twine --skip-existing`** added, making the publish step idempotent for partial uploads / non-bumped pushes. Verified it does not weaken the release gate: `release-image.yml`'s `verify-published` still fails closed for a tag whose version was never actually published.
- **`release-image.yml` operator guidance** corrected (`Publish Package to PyPI (prod)` / `publish-main.yml`).
- **`publish-dev.yml` brought to parity** — `permissions: contents: read`, `checkout@v4`, `setup-python@v5`.
- **`setup-python@v5`** in both files.
- **`cancel-in-progress: false`** in both files (behavior-neutral, now honest).

Core change remains correct: explicit prod-PyPI target, least-privilege permissions, honest step name. Nice work on the follow-up. 🚀

PR #449, opened 2026-08-05
fix(#441): preflight-reject a time_column absent from the CSV header

divyasinghds:
## What

Preflight-reject a configured `time_column` that isn't in the CSV header, so a
time-series ingest fails fast with an actionable message instead of silently
ingesting every row. Fixes #441.

## Why

`time_series_forecasting` (and `time_series_classification`) hardcode the fixed
`timestamp` column in their validators and never consume the config's
`time_column`. So a manifest with `time_column: nonexistent_col` was parsed into
`ResolvedConfig`, then **ignored** — validation "passed" against the real
`timestamp` column and all rows landed. A silent accept that only surfaces later,
as undefined ordering, in training (repro: prod job `ingest-job-af7196f64447`,
10 rows landed).

## How

A single, category-agnostic preflight check that runs before any DB writes:

- **`preflight.check_time_column(source, time_column, csv_options)`** — when
  `time_column` is set, resolves it against the CSV header (tokenized with the
  run's dialect via `read_dialect_kwargs`, matched case-/whitespace-insensitively
  via `resolve_column`, exactly like the ingest read path). On a miss it raises
  `time_column '<name>' not found in CSV columns: [...]`. No-op when unset or for
  non-CSV / missing sources (JSON columns are validated downstream).
- **`base.py`** — stores `self.time_column` (a plain attribute, *not* bridged into
  `file_options`, so it never rides the registration `meta_data` wire payload) and
  calls the check in `validate_data`, right after the encoding preflight.
- **`csv_ingestor.py` / `json_ingestor.py` / `run.py`** — thread `time_column`
  from the resolved config to the ingestor.

This covers all three time-series categories at one site:
`time_to_event_prediction` already rejected a missing column via
`TimeToEventValidator` and still does (now caught one step earlier, with the same
uniform message); TSF/TSC were previously silent and now reject.

## Tests

- `preflight.check_time_column`: rejects missing column, accepts a present one
  (case-insensitive), no-ops when unset/non-CSV, honors a non-comma dialect.
- `validate_data` end-to-end: a TSF-shaped CSV missing the configured column
  aborts before validators run.
- `run.py` wiring: a top-level `time_column` reaches the ingestor kwarg;
  image_classification threads `None`.

Full suite: 1915 passed, 1 xfailed.

## Note on #442

#442 (v0.5.7 image rejects `time_series_classification`) is **not a code bug** —
develop is already at 0.8.2 with TSC in the schema enum, the registry, and every
dispatch site, guarded by `test_schema_validation` + `test_category_congruence`.
That drift was against the stale deployed image and is resolved by cutting/rolling
a new image, so it's out of scope for this PR.

---

> [!NOTE]
> **Medium Risk**
> Changes fail-fast validation on the core ingest path for time-series categories; misconfigured manifests that previously ingested silently will now error, which is intended but affects production configs.
> 
> **Overview**
> Fixes **#441**: manifests with a top-level `time_column` on `time_series_forecasting` / `time_series_classification` used to be ignored while rows still ingested ordered by the platform-fixed **`timestamp`** column.
> 
> **Preflight:** Adds `preflight.check_time_column` and runs it from `BaseIngestor.validate_data` (before validators/DB writes). For categories in `FIXED_TIME_COLUMN_BY_CATEGORY`, any configured `time_column` that does not resolve to the fixed name (case/whitespace-insensitive) raises a clear `ValueError`—including typos and real columns that are not `timestamp`. `time_to_event_prediction` and other categories are unchanged (no-op when there is no fixed column).
> 
> **Config wiring:** `conventions.resolve` bridges top-level `time_column` into `file_options` for the whole time-series family (TSF, TSC, TTE) via `_TIME_COLUMN_CATEGORIES`, so the preflight sees it without duplicating it under `spec.file_options`. `time_column` is added to internal meta keys so it does not leak on registration `meta_data`.
> 
> Bumps package version to **0.8.3**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d7b7a6d65bca6fa56005d831dd62f387fcb0c74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/cli/conventions.py
+++ b/tracebloc_ingestor/cli/conventions.py
@@ -101,6 +101,14 @@ TIME_TO_EVENT_CATEGORIES: FrozenSet[str] = frozenset(
     }
 )
 
+# Categories that accept a top-level `time_column` and bridge it into
+# file_options (step 7a): time_to_event_prediction consumes it; the two
+# time-series categories only get it there so preflight can reject a value that
+# won't be honored (#441). Non-time-series categories ignore the field.
+_TIME_COLUMN_CATEGORIES: FrozenSet[str] = (
+    TIME_SERIES_CATEGORIES | TIME_TO_EVENT_CATEGORIES
+)
+
 MLM_CATEGORIES: FrozenSet[str] = frozenset(
     {
         TaskCategory.MASKED_LANGUAGE_MODELING,
@@ -213,7 +221,9 @@ class ResolvedConfig:
 
     # ----- data_id -----
     unique_id_column: Optional[str] = None  # None ⇒ strategy below decides
-    data_id_strategy: str = "content_hash"  # content_hash (default, #350) | uuid (column ⇒ unique_id_column)
+    data_id_strategy: str = (
+        "content_hash"  # content_hash (default, #350) | uuid (column ⇒ unique_id_column)
+    )
 
     # ----- Pass-through to ingestors -----
     annotation_column: Optional[str] = None
@@ -329,14 +339,20 @@ def resolve(config: Dict[str, Any]) -> ResolvedConfig:
     spec_file_options = (config.get("spec") or {}).get("file_options") or {}
     resolved.file_options = {**_default_file_options_for(category), **spec_file_options}
 
-    # 7a. For time_to_event_prediction the validator (TimeToEventValidator)
-    #     reads `time_column` from file_options. Bridge the top-level field
-    #     so the validator gets it without customers having to repeat the
-    #     value in spec.file_options. ``setdefault`` so an explicit
-    #     spec.file_options.time_column (the advanced override) wins over
-    #     the documented top-level shorthand, consistent with how every
-    #     other spec.file_options key behaves.
-    if category == TaskCategory.TIME_TO_EVENT_PREDICTION and resolved.time_column:
+    # 7a. Bridge the top-level `time_column` into file_options for the
+    #     time-series family so the ingestor sees it without customers having to
+    #     repeat the value in spec.file_options. ``setdefault`` so an explicit
+    #     spec.file_options.time_column (the advanced override) wins over the
+    #     top-level shorthand, consistent with how every other spec.file_options
+    #     key behaves.
+    #       - time_to_event_prediction: TimeToEventValidator consumes it (the
+    #         column is user-configurable).
+    #       - time_series_forecasting / _classification: the column is FIXED to
+    #         the physical `timestamp` and a custom value is never honored, so
+    #         BaseIngestor.validate_data preflights it and REJECTS anything else
+    #         (a typo or a real-but-wrong column) rather than silently ordering
+    #         by `timestamp` anyway (#441).
+    if category in _TIME_COLUMN_CATEGORIES and resolved.time_column:
         resolved.file_options.setdefault("time_column", resolved.time_column)
 
     # 7b. Bridge top-level `target_size` (any image category — customer
--- a/tracebloc_ingestor/ingestors/base.py
+++ b/tracebloc_ingestor/ingestors/base.py
@@ -41,6 +41,7 @@ from .table_lock import TableLock
 # (and their None/unknown -> False semantics are preserved).
 from ..modalities.registry import (
     FILE_BEARING_CATEGORIES as _FILE_BEARING_CATEGORIES,
+    FIXED_TIME_COLUMN_BY_CATEGORY,
     NLP_CATEGORIES as _NLP_CATEGORIES,
     REGISTRY as _MODALITY_REGISTRY,
     TABULAR_FAMILY_CATEGORIES as _TABULAR_FAMILY_CATEGORIES,
@@ -240,7 +241,6 @@ class BaseIngestor(ABC):
         file_options: Optional[Dict[str, Any]] = None,
         label_policy: str = label_policy_module.PASSTHROUGH,
         data_id_strategy: str = "content_hash",
-        time_column: Optional[str] = None,
     ):
         """Initialize the base ingestor.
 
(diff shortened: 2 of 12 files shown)


tracebloc_ingestor/ingestors/base.py
```diff
@@ -552,6 +562,16 @@ def validate_data(self, source: Any) -> bool:
         # Catch it once here with a clear, actionable message.
         preflight.check_csv_encoding(source)
 
+        # Pre-flight: a configured time_column that isn't in the CSV header is
+        # otherwise silently accepted — the time-series validators fall back to
+        # the fixed ``timestamp`` column and the rows land ordered by nothing
+        # (#441). Reject the typo up front, before any DB writes. Reads the
+        # header with the run's CSV dialect (getattr: only CSVIngestor carries
+        # csv_options).
+        preflight.check_time_column(
```
saadqbal, line 573: This guards the *typo* case but not the one #441 is really about. For TSF/TSC `time_column` is never consumed — ordering is always the fixed physical `timestamp` column (registry.py:267, and the schema even documents `time_column` as time_to_event only). So `time_column: some_real_but_wrong_col` (a column that IS in the header, just not `timestamp`) sails through `resolve_column` here and still lands rows ordered by `timestamp` — the same silent-accept as the original repro, just a different subcase. Only `time_column: nonexistent` gets caught.

If we want to actually close the silent-accept for TSF/TSC, the honest fix is to reject any `time_column` that isn't `timestamp` for those categories (or thread it into the validators). As-is the preflight validates a decorative field and can give false confidence that a custom time_column is being honored.

tracebloc_ingestor/ingestors/preflight.py
```diff
@@ -103,6 +106,61 @@ def check_csv_encoding(source: Any) -> None:
...
+    try:
+        header = pd.read_csv(path, nrows=0, **read_dialect_kwargs(csv_options))
+    except Exception:
+        # A malformed / unreadable CSV surfaces with clearer messages through
+        # check_csv_encoding and the data validators; don't fail on an
+        # unrelated parse error here (nor mask the real cause).
+        return
+
+    columns = list(header.columns)
+    if resolve_column(columns, time_column) is None:
```
saadqbal: Small semantics split worth noting: this resolves case-/whitespace-insensitively, but `TimeToEventValidator` matches exactly (`time_column not in df.columns`, time_to_event_validator.py:106). So for TTE a `time_column: Time` against header `time` passes here and is then rejected downstream by the validator with its *old* message — not the "caught one step earlier, uniform message" the PR describes. Data-safe (still rejected), just not the uniform path the description claims.

saadqbal (commented):
Solid, well-tested fix for the typo case. One thing that keeps it from being a full close on #441, plus a small semantics note — inline. Not blocking, but the first one is worth a look before merge.

tracebloc_ingestor/ingestors/base.py
divyasinghds: Fixed in 23d7b5b. The preflight is now category-aware: it rejects ANY configured time_column that isn't the category's fixed column for TSF/TSC — so the real-but-wrong `temp_c` case is caught too, not just a nonexistent name. The fixed column is a single source of truth in `registry.FIXED_TIME_COLUMN_BY_CATEGORY` (TSC derived from its grouping trait). It no longer validates against the header (decorative field → false confidence, as you noted).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tracebloc_ingestor/ingestors/preflight.py
divyasinghds: Moot now (23d7b5b): time_to_event_prediction is no longer preflighted here — the check runs only for the fixed-timestamp categories (TSF/TSC). TimeToEventValidator stays the sole gate for TTE with its exact match, so there's no case-insensitive-vs-exact split anymore.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tracebloc_ingestor/ingestors/base.py
```diff
@@ -552,6 +562,16 @@ def validate_data(self, source: Any) -> bool:
...
+            source, self.time_column, getattr(self, "csv_options", {})
+        )
```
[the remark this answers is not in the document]
divyasinghds: Resolved by the redesign in 23d7b5b: TTE is no longer preflighted at all (only the fixed-timestamp categories are), so the top-level-vs-`spec.file_options.time_column` precedence divergence can't occur here — TimeToEventValidator remains the exact gate for TTE.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tracebloc_ingestor/cli/conventions.py
```diff
@@ -329,14 +339,20 @@ def resolve(config: Dict[str, Any]) -> ResolvedConfig:
...
+    #     top-level shorthand, consistent with how every other spec.file_options
+    #     key behaves.
+    #       - time_to_event_prediction: TimeToEventValidator consumes it (the
+    #         column is user-configurable).
+    #       - time_series_forecasting / _classification: the column is FIXED to
+    #         the physical `timestamp` and a custom value is never honored, so
+    #         BaseIngestor.validate_data preflights it and REJECTS anything else
+    #         (a typo or a real-but-wrong column) rather than silently ordering
+    #         by `timestamp` anyway (#441).
+    if category in _TIME_COLUMN_CATEGORIES and resolved.time_column:
```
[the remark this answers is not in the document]
divyasinghds: Fixed in 3d7b7a6: added `time_column` to `_META_DATA_INTERNAL_KEYS` so it's stripped from `_meta_data_payload` like the other internal bridges. It's bridged only to configure the preflight validator, not as dataset metadata — and nothing consumes `meta_data.time_column` (the survival facts that do belong on the wire, `time_unit`/`event_indicator`, still ride `attributes`). This also closes the pre-existing TTE leak. Extended `test_meta_data_payload_strips_scalar_alignment_bridges` to cover it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

saadqbal (approved):

PR #466, opened 2026-08-10
sec(#1425): publish to PyPI via Trusted Publishing (OIDC)

saadqbal:
## What

Replace the long-lived `PYPI_API_TOKEN` repo secret with **PyPI Trusted Publishing (OIDC)** in `publish-main.yml`. The runner mints a short-lived OIDC token (`id-token: write`) that PyPI exchanges for an upload scope — no static token in this **public** repo's secrets. Swaps the `twine upload` step for the SHA-pinned `pypa/gh-action-pypi-publish@dc37677` (v1.14.2); `skip-existing: true` keeps the idempotency the old `--skip-existing` gave us.

Closes the `data-ingestors` half of **backend#1425**.

## ⛔ Blocked — do not merge yet (draft on purpose)

A **trusted publisher must be registered on PyPI first**, or the next push to `main` fails to publish:
- Project: `tracebloc_ingestor`
- Owner: `tracebloc` · Repo: `data-ingestors` · Workflow: `publish-main.yml` · Environment: *(none)*

@LukasWodka owns that registration per backend#1425. Once it's live, mark ready → then `PYPI_API_TOKEN` can be deleted + revoked.

## Verification

- [ ] Trusted publisher registered on PyPI (precondition above)
- [ ] Repo CI green on this PR (local YAML lint not run — pyyaml/actionlint unavailable in my env; relying on CI + the diff)
- [ ] After merge + a real release: confirm OIDC publish succeeds, then delete `PYPI_API_TOKEN` and `GH_PACKAGES_TOKEN` repo secrets

### DIFF
--- a/.github/workflows/publish-main.yml
+++ b/.github/workflows/publish-main.yml
@@ -11,10 +11,12 @@ concurrency:
   # cancel; a publish in flight must not be cancelled anyway. State it plainly.
   cancel-in-progress: false
 
-# Least privilege: this job publishes to PyPI with PYPI_API_TOKEN, not with the
-# workflow token, so it needs nothing beyond reading the checked-out source.
+# Least privilege. Publishing uses PyPI Trusted Publishing (OIDC), so the job
+# needs id-token to mint the short-lived OIDC credential plus read access to the
+# checked-out source -- no long-lived PyPI token in repo secrets (backend#1425).
 permissions:
   contents: read
+  id-token: write
 
 jobs:
   publish:
@@ -58,17 +60,17 @@ jobs:
           python -c "from tracebloc_ingestor.cli.run import _load_schema; print(_load_schema()['title'])"
 
       - name: Publish to production PyPI
-        # This step was named "Publish to GitHub Packages", but `twine upload`
-        # with no --repository-url targets upload.pypi.org -- so every push to
-        # the prod branch published to PRODUCTION PyPI while claiming otherwise
-        # (backend#1427). The target is now explicit and the step is named for
-        # what it does. --skip-existing makes the step idempotent: a re-run
-        # after a partial upload (wheel landed, sdist hit a transient error), or a
-        # push that did not bump the version, becomes a green no-op instead of a
-        # permanent HTTP 400 "File already exists" with no clean re-run path.
-        run: |
-          python -m pip install twine
-          twine upload --skip-existing --repository-url https://upload.pypi.org/legacy/ dist/*
-        env:
-          TWINE_USERNAME: '__token__'
-          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+        # PyPI Trusted Publishing (OIDC): the runner mints a short-lived OIDC
+        # token that PyPI exchanges for an upload scope, so there is no long-lived
+        # PYPI_API_TOKEN in repo secrets (backend#1425). The default repository
+        # target is upload.pypi.org (production PyPI). skip-existing keeps the step
+        # idempotent -- a re-run after a partial upload (wheel landed, sdist hit a
+        # transient error), or a push that did not bump the version, is a green
+        # no-op instead of a permanent HTTP 400 "File already exists".
+        #
+        # PRECONDITION: a trusted publisher must be registered on PyPI for project
+        # `tracebloc_ingestor` (owner tracebloc, repo data-ingestors, workflow
+        # publish-main.yml, no environment) before this step can succeed.
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
+        with:
+          skip-existing: true


saadqbal:
@LukasWodka — taking this out of draft. It has been sitting since 2026-08-11 for no reason other than nobody marked it ready, and it removes a long-lived publishing credential, so it should not keep waiting.

Brought current first: merged `origin/develop` (it was 48 commits behind), pushed as `af2b890`. Clean merge, and I checked the substance survived it rather than assuming:

- `id-token: write` present, `pypa/gh-action-pypi-publish` pinned to a SHA (`dc37677b2e1c63e2034f94d8a5b11f265b73ba33`, v1.14.2)
- **no live `${{ secrets.PYPI_API_TOKEN }}` reference anywhere in the repo** — 0 files; the only remaining mention is the explanatory comment
- `publish-main.yml` still parses as YAML
- `publish-dev.yml` does no PyPI upload, so `publish-main.yml` was the only site — the class is closed, not just this instance

Still open as backend#1425. One thing needs doing outside the PR and I cannot do it: **PyPI's side must have a Trusted Publisher configured** for this repo/workflow/environment before the first publish, or the upload fails at the OIDC exchange. If that is not already set up, this should not merge until it is — worth confirming before you approve rather than finding out on a release.

saqlainsyed007 (approved):
Approve — the Trusted Publishing (OIDC) migration is correct and complete.

- `id-token: write` is granted, and since `publish-main.yml` has a single `publish` job the workflow-level block is effectively job-scoped (job-level would be a marginal style improvement, not required).
- The `twine upload` step and `TWINE_USERNAME`/`TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}` are fully removed — no token fallback left. Deleting/revoking the repo secret is correctly a post-merge manual step.
- Trigger stays `push: branches: [main]`, so a PR can't publish.
- `pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33` is SHA-pinned; verified it dereferences to the `v1.14.2` tag. `skip-existing: true` preserves the old `--skip-existing` idempotency.
- No `environment:` is intentional and documented, and matches the PyPI trusted-publisher registration (no environment).
- No secret added; gitleaks green.

One non-code gate before merge: the trusted publisher for `tracebloc_ingestor` must be registered on PyPI first (backend#1425), or the next push to `main` fails to publish. The PR body still says "do not merge yet (draft on purpose)" but the PR is no longer a draft — please confirm the registration is live (and refresh that body text) before merging.

LukasWodka (approved):
Approving at `af2b890e80218bca21550dd9e354ab20edad9ee6`. Small diff, meaningful surface — I checked the parts that decide whether it's actually safer rather than just differently written.

**The action pin is correct, and I verified it the hard way.** `pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2` — `v1.14.2` is an *annotated* tag, so the ref API returns the tag object `a892a5a6…`, which does not match the pin at first glance. Dereferencing the tag gives commit `dc37677b2e1c63e2034f94d8a5b11f265b73ba33`, exactly what's pinned. The comment and the SHA agree; I mention it only because the annotated-tag indirection makes this look like a mismatch to anyone spot-checking it, including a future audit script.

**The token is genuinely gone, not just moved.** `TWINE_USERNAME`/`TWINE_PASSWORD` and the `secrets.PYPI_API_TOKEN` reference are removed; the only remaining occurrence of that string at head is the explanatory comment on line 65. `main` has two occurrences, this head has one, and it's prose.

**Permissions are right.** `contents: read` retained, `id-token: write` added — that's the minimum Trusted Publishing needs, and `id-token` is genuinely required rather than defensive.

`skip-existing: true` is carried over, which keeps the idempotence the previous `--skip-existing` bought: a re-run after a partial upload, or a push that didn't bump the version, stays a green no-op instead of a permanent 400.

Also worth saying: replacing the hand-rolled `twine upload --repository-url https://upload.pypi.org/legacy/` with the official action removes the footgun that produced backend#1427 in the first place — a step named for one target that silently uploaded to another. The default target is now the action's, not a URL someone has to keep correct.

## Three follow-ups, none blocking

**1. Delete the `PYPI_API_TOKEN` repo secret once this lands.** This is the actual point of the change and the step most often forgotten: the workflow no longer *uses* the token, but if the secret stays in the repo it remains a long-lived PyPI credential that any workflow with `secrets` access can read. The diff can't do this — it needs a settings change — so it's worth doing deliberately rather than assuming the removal from YAML finished the job.

**2. Confirm the trusted publisher is registered before merging.** Your comment documents the precondition (project `tracebloc_ingestor`, owner `tracebloc`, repo `data-ingestors`, workflow `publish-main.yml`, no environment). If it isn't registered when this merges, the next push to `main` fails at the publish step — noisy rather than dangerous, but it fails *after* the build and version bump, which is an awkward place to discover it.

**3. Consider binding the publish job to a GitHub Environment.** Trusted Publishing supports it, and it's the one hardening this change leaves on the table: with `on: push: branches: [main]`, any merge to `main` publishes to production PyPI with no gate. An environment would let you add required reviewers or a wait timer on the publish specifically. Not a regression — that's the behaviour today — and registering with "no environment" as you've done is the simpler starting point. Worth a follow-up ticket rather than a change here, since adding one later means re-registering the publisher on PyPI.

One cosmetic note: `permissions:` is workflow-level, and with a single `publish` job that's equivalent to scoping it to the job. If a second job is ever added to this file it would inherit `id-token: write` for free, so job-level scoping is slightly more robust against future edits. Not worth changing on its own.

CI green, no open threads, `CLEAN/MERGEABLE`, and @saqlainsyed007 has approved.

PR #468, opened 2026-08-10
sec(#1528): require DB_USER/DB_PASSWORD, drop the edgeuser fallback (D10) [blocked on S2]

divyasinghds:
## What

Remove the hardcoded `edgeuser` / `Edg9@Tr@ce` fallback for `DB_USER` / `DB_PASSWORD` in `config.py`. Both are now **required from env** and fail fast (via a new `_require_env` helper mirroring `_as_int`) with a message that names the variable and points at `tb_ingest` — instead of silently authenticating to MySQL as the root-equivalent `edgeuser`.

This is the **data-ingestors slice** of backend#1528 (RFC-0003 D10 close-out: retire `edgeuser` root-equivalence, mint `tb_ingest` + `tb_meta`).

## Why

The ingestor already reads `DB_USER`/`DB_PASSWORD` from env (`database.py` `_create_engine`); the only thing pinning it to `edgeuser` is the config-level default. backend#1528 mints a dedicated least-privilege `tb_ingest` identity (`CREATE, ALTER, SELECT, INSERT, UPDATE, DELETE` on `training_test_datasets.*`) and has jobs-manager inject it into each spawned ingestion Job. Once that lands, this fallback is dead — and while it lives, a misconfigured Job silently connects as root-equivalent `edgeuser` rather than failing loudly.

## Changes

- **`tracebloc_ingestor/config.py`** — `DB_USER`/`DB_PASSWORD` require env; new `_require_env` fail-fast helper; `===== Database =====` comment rewritten to state the new contract (`DB_HOST`/`DB_PORT`/`DB_NAME` keep their conventional defaults; `DB_USER`/`DB_PASSWORD` do not). Lazy `@property`, so only the DB path triggers the check.
- **`tests/test_no_hardcoded_secrets.py`** — `Edg9@Tr@ce` and `edgeuser` moved from the allowlist into the regression guard; they can no longer reappear in package source, even in a comment.
- **`tests/test_config_lazy.py`** — fail-fast (unset → `ValueError` naming the var), env-flow, and override tests.
- **`tests/test_database.py`, `tests/test_per_ingestion_tables.py`** — the two mocked-engine `Database` fixtures now pass throwaway creds (matches the existing pattern in `test_time_series_classification.py`).

## Verification

- `pytest`: **1927 passed, 1 xfailed** (pre-existing).
- `ruff check`: clean.

## ⚠️ Release constraint — DRAFT until S2

This must **not** reach any fleet until backend#1528 **S2** (client-runtime `submit_ingestion_run` stamping `DB_USER`/`DB_PASSWORD=tb_ingest` onto spawned ingestion Jobs) is deployed there. Sequence:

```
S1 mint  →  S2 inject env  →  (this PR) drop fallback  →  S3 REVOKE edgeuser
```

Pre-S2, jobs-manager sets no DB creds, so an ingestor image built from this branch would hit the new fail-fast at init. S1 is merged but not enabled anywhere; **S2 is not yet built**. Keeping this a draft until S2 has deployed fleet-wide.

Refs: backend#1528 (D10 close-out), client-runtime#290 / client#643 (S1).

---

> [!NOTE]
> **High Risk**
> Misconfigured Jobs without injected DB creds will fail at Database init; fleet rollout must follow backend#1528 S2 (jobs-manager injecting tb_ingest) before shipping this image.
> 
> **Overview**
> **Requires `DB_USER` and `DB_PASSWORD` from the environment** — the hardcoded `edgeuser` / `Edg9@Tr@ce` defaults are removed (backend#1528 D10). A new `_require_env` helper raises a clear `ValueError` naming the missing variable and pointing at per-Job `tb_ingest` injection instead of silently using the legacy MySQL identity.
> 
> `Config.validate()` stays **auth-scoped** (backend token / client creds only); DB credentials are enforced on first property access when `Database()` initializes. Host/port/database name defaults are unchanged.
> 
> Regression guards now forbid `Edg9@Tr@ce` and `edgeuser` string literals in package source. Tests cover fail-fast, env flow, and overrides; mocked `Database` fixtures supply throwaway creds. Version **0.8.5**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 820e9f5410c7780de0de319d59680d4214d986e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/__init__.py
+++ b/tracebloc_ingestor/__init__.py
@@ -28,7 +28,7 @@ from .validators import (
 # Single source of truth for the package version. setup.py parses this literal
 # (see _read_version in setup.py) so the two can't drift again (#175). Bump here
 # only — setup.py picks it up automatically.
-__version__ = "0.8.4"
+__version__ = "0.8.5"
 
 __all__ = [
     "Config",
--- a/tracebloc_ingestor/__init__.py
+++ b/tracebloc_ingestor/__init__.py
@@ -28,7 +28,7 @@ from .validators import (
 # Single source of truth for the package version. setup.py parses this literal
 # (see _read_version in setup.py) so the two can't drift again (#175). Bump here
 # only — setup.py picks it up automatically.
-__version__ = "0.8.5"
+__version__ = "0.8.4"
 
 __all__ = [
     "Config",
--- a/tests/test_api_client_auth.py
+++ b/tests/test_api_client_auth.py
@@ -11,10 +11,10 @@ Covers all three boot paths required by the #43 acceptance criteria:
 A fourth path — ``CLIENT_ENV=local`` — is also exercised because it bypasses
 both validation and the network entirely.
 
-Database credentials (``DB_USER`` / ``DB_PASSWORD``) are also part of the
-validation surface as of backend#1528: the root-equivalent ``edgeuser``
-fallback was removed, so a real run requires the ``tb_ingest`` identity
-jobs-manager injects per-Job. See `Config.validate()`.
+Database credentials (``DB_USER`` / ``DB_PASSWORD``) are required for a real
+run as of backend#1528 (the ``edgeuser`` fallback was removed — jobs-manager
+injects ``tb_ingest`` per-Job), but they are NOT part of this auth-scoped
+``validate()``; they fail fast at ``Database()`` init instead.
 """
 
 from __future__ import annotations
@@ -33,16 +33,13 @@ def _make_config(**overrides) -> Config:
 
     Default: a valid prod-like config with BACKEND_TOKEN set so ``validate()``
     passes; tests override only the auth fields they care about. DB creds are
-    now part of the validation surface (backend#1528 removed the edgeuser
-    fallback — jobs-manager injects tb_ingest per-Job), so the defaults include
-    them; a test can drop them via ``DB_USER=None`` / ``DB_PASSWORD=None``.
+    not part of ``validate()`` (it's auth-scoped) — they fail fast at
+    ``Database()`` init instead — so they're left unset here.
     """
     defaults = dict(
         BACKEND_TOKEN="test-token-abc",
         CLIENT_USERNAME=None,
         CLIENT_PASSWORD=None,
-        DB_USER="tb_ingest",
-        DB_PASSWORD="pw",
         EDGE_ENV="prod",
     )
     defaults.update(overrides)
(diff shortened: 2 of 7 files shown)


tracebloc_ingestor/__init__.py
```diff
@@ -28,7 +28,7 @@
 # Single source of truth for the package version. setup.py parses this literal
 # (see _read_version in setup.py) so the two can't drift again (#175). Bump here
 # only — setup.py picks it up automatically.
-__version__ = "0.8.4"
+__version__ = "0.8.5"
```
saadqbal, line 31: Drop this bump. Hand-bumping a version file is the release train's job, not a feature PR's — the standards doc says so outright ("Never hand-cut a `v*` tag, hand-bump a version file, or publish an artifact"), and every bump in this repo's history is its own `chore(release):` commit. `publish-main.yml` ships this literal to real PyPI on the train-owned merge to `main`, so it's the train's to set.

Also a collision risk: the PR is pinned open until backend#1528 S2, so by the time it lands 0.8.5 is likely already taken by whatever the train cut in the meantime.

saadqbal (commented):
Clean, well-scoped change and the fail-fast tests are thorough — no correctness issues. Two things before it goes in:

- Version bump (inline): drop it. Hand-bumping `__version__` in a feature/sec PR isn't ours to do — every bump in the log is its own `chore(release):` commit, and the standards doc is explicit ("Never hand-cut a `v*` tag, hand-bump a version file, or publish an artifact"). Extra sharp here because the PR is pinned open until S2: a hardcoded 0.8.5 will likely collide with whatever the train cuts before this lands.
- `Config.validate()`'s docstring is now false. It still says DB creds aren't validated because "the bundled MySQL container ships with fixed credentials that the ingestor defaults match … connection convention, not a secret … forcing customers to set them in env vars adds friction with no security benefit" — that's the exact rationale this PR reverses. Update it in the same PR. And worth a thought: now that DB creds are required for any real run, should they join the boot-time check in `validate()` rather than only surfacing deep in `_create_engine`?

Minor: the wholesale black-reformat of test_database.py / test_config_lazy.py buries the two-line fixture change under ~80 lines of whitespace — ideally its own formatting PR (refactors separate from behavior).

tracebloc_ingestor/__init__.py
divyasinghds: Dropped — reverted `__version__` to 0.8.4 in 70baa29. Agreed the train owns version bumps (and the collision risk while this is pinned open until S2). The "package change ⇒ __version__ bump" gate will sit red while the PR is held; the train sets the version at the train-owned merge.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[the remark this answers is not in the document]
divyasinghds: Not auto-resolving — this is a live disagreement between two maintainers. The 0.8.5 bump is intentional: @cryptoalak re-added it in 820e9f5 arguing the train's `read_version`/`latest_final` regression guard would refuse to re-tag 0.8.4 at promotion (so the change can't reach a released image without it). That conflicts with @saadqbal's review here asking to drop the hand-bump (train owns versions; collision risk while the PR is held). Leaving the bump in place and deferring the version-policy call to @saadqbal + @cryptoalak — see the PR comment below.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

saadqbal (approved):

aptracebloc (approved):

shujaatTracebloc (approved):
🚫 DO NOT MERGE until backend#1528 S2 has shipped fleet-wide — see the release-safety note below. Code-wise: approving.

The change is clean and correct: `_require_env` removes the hardcoded edgeuser/Edg9@Tr@ce MySQL fallback and fails fast on a missing DB_USER/DB_PASSWORD with a message that names the variable and points at per-Job tb_ingest. It mirrors `_as_int`, stays out of the auth-scoped validate() (fails at Database() init instead), and the new parametrized test covers the no-fallback requirement. CI green, and it already carries saadqbal + aptracebloc approvals.

Two release-safety flags (not code):
1. Per your own PR body the sequence is S1 mint → S2 inject env → (this) drop fallback → S3 revoke, and 'S2 is not yet built.' Pre-S2, jobs-manager injects no DB creds, so an ingestor image built from this branch hits the new fail-fast at init and breaks ingestion fleet-wide. This must not merge/ship before S2 is deployed everywhere.
2. The body says 'Keeping this a draft until S2 has deployed fleet-wide,' but the PR is currently NOT a draft and has two approvals + green CI — i.e. it looks mergeable. Recommend converting it back to draft, or adding a do-not-merge / blocked label, so it can't be merged early by mistake.

PR #477, opened 2026-08-11
fix(transfer): make the dest dir removable by the teardown pod (different uid, no shared group)

shujaatTracebloc:
Closes #476.

## The bug

`data delete` dropped the table, then failed to remove the files:

```
rm: can't remove '/data/shared/<table>/<file>.jpg': Permission denied
```

The dataset is left half-deleted — gone from the database, still on disk — and the error's
"re-run the delete" advice can't work, because the table it would look for is already gone.

## Root cause

Removing a file needs write+execute on its **parent directory**, not on the file. The pods that
create and delete these trees share no identity:

| Actor | Identity |
|---|---|
| ingestion Job (creates dir + files) | uid 65534 (`nobody`), or `HOST_UID` |
| CLI staging/teardown pod (removes them) | uid 65532, `fsGroup` 65532 |
| the shared mount itself | chowned `1000:1000` + setgid → a dir created underneath inherits group **1000** |

`DEST_DIR_MODE` was `0o2775` — setgid + **group**-writable — and its comment said the teardown
group could reclaim the tree. It couldn't: the inherited group is 1000, the teardown pod carries
65532. `fsGroup` normally fixes precisely this by making kubelet recursively `chgrp` the volume,
but **kubelet ignores `fsGroup` on hostPath volumes**
([kubernetes#138411](https://github.com/kubernetes/kubernetes/issues/138411)) — the layout every
installer-provisioned cluster uses.

Why it shipped green: the existing test asserted the mode **bits** and never that the group
matched the teardown's identity, so a mechanism that cannot work passed its own test.

**Second, independent defect.** The duplicate validator creates the *same* `DEST_PATH` with a bare
`mkdir` and runs **before** the transfer, and `_ensure_dest_dir` deliberately does not re-chmod a
directory that already exists. So on every run where validation created the directory first, the
mode fix was dead code and the tree stayed at the umask default (0755).

Either defect alone leaves the bug reachable, so both are fixed here.

## The change

- **New `tracebloc_ingestor/utils/fs.py`** holds the single rule. `file_transfer` re-exports
  `DEST_DIR_MODE` and keeps `_ensure_dest_dir` as a thin delegation, so existing callers and tests
  are untouched.
- **The validator uses the shared helper**, so the mode applies whichever side creates the dir.
- **Mode widens to `0o2777`.** Nothing at ingest time can know the teardown's uid, and the parent
  mount is already world-writable by chart design — so this is not a wider grant than the tree it
  sits in.

**Deliberately unchanged:** a pre-existing directory is still never re-chmod'd, so an operator's
mode is not overridden and a genuinely unwritable destination still surfaces downstream. Trees
left by an older ingestor keep their old mode — repairing those is a node-side one-off, not a
silent chmod of data the ingestor didn't write.

## Verification

- `pytest` — **1923 passed, 1 xfailed** (3 new tests).
- **Each new test was proved to bite**: reverting the mode to `0o2775` fails
  `test_ensure_dest_dir_is_world_writable_not_just_group_writable`, and restoring the validator's
  bare `mkdir` fails `test_duplicate_validator_creates_dest_reclaimable_too`. A third test pins
  both sites to one mode definition so the two copies can't drift apart again.
- `black` + `ruff` clean.

Found while debugging a Windows install where the first ingest and then the first delete both
failed on hostPath permissions. The ingest half is fixed in the client installer; this is the
delete half.

---

> [!NOTE]
> **Medium Risk**
> Changes shared PVC directory permissions and delete teardown behavior; scope is narrow and well-tested, but mistakes could leave undeletable data or widen per-dir writability on the mount.
> 
> **Overview**
> Fixes **`data delete`** leaving datasets half-removed on disk when the teardown pod (different uid, no shared group with the ingest job on hostPath) cannot delete files under the table directory.
> 
> **Centralizes destination-dir creation** in new `tracebloc_ingestor/utils/fs.py`: `ensure_reclaimable_dir` and `DEST_DIR_MODE` (**`0o2777`**, setgid + world write/execute) replace the previous **`0o2775`** group-writable assumption. `file_transfer._ensure_dest_dir` now delegates there and still re-exports the constants for existing callers/tests.
> 
> **`DuplicateValidator`** no longer uses a bare `mkdir` when it creates `DEST_PATH` first (validation runs before transfer, and existing dirs are not re-chmodded), so the reclaimable mode applies whichever path creates the directory.
> 
> Adds regression tests for world-writable mode, validator-created dirs, and a single shared mode definition. Package version **0.8.5**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cd556965eb73e10b0e3641796251c3095cd8f2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_file_transfer_failure_modes.py
+++ b/tests/test_file_transfer_failure_modes.py
@@ -170,3 +170,62 @@ def test_segmentation_missing_mask_leaves_no_orphan(dirs):
     )
     assert rec is None
     assert not dest.exists() or not any(dest.iterdir())
+
+
+# --- client#653: the teardown runs as a DIFFERENT uid, so group-write is not enough
+
+
+@pytest.mark.skipif(
+    hasattr(os, "geteuid") and os.geteuid() == 0,
+    reason="root's created dirs ignore umask/perms nuances",
+)
+def test_ensure_dest_dir_is_world_writable_not_just_group_writable(tmp_path):
+    """Removing a file needs write on its PARENT DIR, and the pods that create
+    and delete these trees never share an identity: the ingestion Job runs as
+    65534 (or HOST_UID), the CLI teardown pod as 65532/fsGroup 65532, and the
+    chart chowns the shared mount to 1000:1000 with setgid -- so a dir created
+    underneath inherits group 1000, never 65532.
+
+    Group-write therefore could not work, and fsGroup cannot rescue it because
+    kubelet ignores fsGroup on hostPath volumes (kubernetes#138411) -- the layout
+    every installer-provisioned cluster uses. The observed failure was
+    "rm: can't remove '/data/shared/<table>/<file>': Permission denied" with the
+    table already dropped, leaving a half-deleted dataset.
+    """
+    dest = tmp_path / "storage" / "tbl"
+    file_transfer._ensure_dest_dir(str(dest))
+    mode = os.stat(dest).st_mode
+    assert mode & 0o002, "other-write must be set: the teardown uid shares no group"
+    assert mode & 0o001, "other-execute must be set, or the dir cannot be traversed"
+
+
+@pytest.mark.skipif(
+    hasattr(os, "geteuid") and os.geteuid() == 0,
+    reason="root's created dirs ignore umask/perms nuances",
+)
+def test_duplicate_validator_creates_dest_reclaimable_too(tmp_path):
+    """The duplicate validator creates the SAME DEST_PATH and runs BEFORE the
+    transfer, so it decides the mode -- and _ensure_dest_dir deliberately does not
+    re-chmod a dir that already exists. While this used a bare mkdir, the mode fix
+    in file_transfer was dead code on every run that validated first: the tree was
+    left at the umask default and the teardown could not remove it (client#653).
+    """
+    from tracebloc_ingestor.validators.duplicate_validator import DuplicateValidator
+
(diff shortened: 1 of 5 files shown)


divyasinghds (approved):
Approving — clean fix for the half-deleted-dataset bug, and the single-source-of-truth refactor is the right call.

Root cause verified: `DEST_DIR_MODE` was 0o2775 (group-write), but the teardown pod (uid 65532) shares no group with the GID 1000 the shared mount inherits, and kubelet ignores fsGroup on hostPath (k8s#138411) — so group-write could never work. Two independent defects, both fixed:
1. mode → 0o2777 (world-write). Justified: no group alignment is possible and the parent mount is already world-writable by chart design, so it's no wider a grant than the tree it sits in.
2. the duplicate validator created the same DEST_PATH first with a bare mkdir (umask 0755) *before* file_transfer's chmod, and _ensure_dest_dir won't re-chmod an existing dir — so the mode fix was dead code on any validate-first run. Now both sites delegate to utils/fs.py.

`test_both_creation_sites_share_one_mode_definition` asserting `is` identity is a good anti-drift guard; root-skip on the perms tests is correct. Version bump gated and green.

Pairs with client#667 (the /data/shared sticky-bit half) — land them together; neither alone fully fixes delete. ✅

PR #487, opened 2026-08-12
fix(#486): apply label.policy at the send boundary, not on the DB write

shujaatTracebloc:
Closes #486

## The bug

`label.policy: bucket` was applied in `RecordProcessor.clean_record` — to the record handed to the batch writer for **INSERT**. `process()` also drops the source `label_column` from the schema-filtered columns, so the raw target was then persisted nowhere: every regression-class dataset (`tabular_regression`, `time_series_forecasting`, `time_to_event_prediction`) stored `sha256(str(value))[:8] % 64` where its training target belonged.

The training client reads that row. A control meant to keep raw targets out of the central backend destroyed the local data instead:

- `time_to_event_prediction` — hard failure. The engine requires the event column ∈ {0,1}; a clean `0`/`1` indicator lands as `56`/`33` and training aborts with `TTE data validation failed: 'label' column must contain only 0/1 binary event indicators. Found unexpected values: ['33', '56']`.
- `tabular_regression`, `time_series_forecasting` — **silent**. A continuous target becomes a bucket id 0–63 and training "succeeds" on meaningless labels.

Both `tracebloc_ingestor/cli/__init__.py:13` ("Apply `label.policy` bucketing in `APIClient.send_batch`") and `utils/label_policy.py`'s own docstring ("the single point where raw labels become bucket IDs **before** `APIClient.send_batch` builds its payload") already described the intended design. No send path implemented it.

## The fix

Move the policy to the boundary it was documented to live at — `APIClient.send_ingest_summary` — and let `RecordProcessor` persist the raw target.

- `label_policy.apply_to_label_counts(counts, policy)` re-keys `{label: row_count}` to bucket ids, **summing** counts of raw values that collide. With 64 buckets collisions are routine; dropping a colliding entry would under-report the dataset's row count to the backend.
- `label_policy.apply_to_samples(samples, policy)` maps the `label` of each outbound `{data_id, label}` preview record, without mutating the caller's list.
- `send_ingest_summary` takes `label_policy` (default `passthrough`) and applies both to the payload — before the local-mock branch too, so a dry run logs the label count that would actually ship.
- `RecordProcessor` no longer accepts a policy at all: a constructor that took one and ignored it is the next version of this bug.

Classification is unchanged — `passthrough` is a no-op, and the default keeps every existing caller's payload byte-identical.

## Notes for the reviewer

- **`data_id` values change for regression-class sources.** The content hash covers the stored record, which now carries the raw label. Re-ingesting the same CSV across this change mints new ids. Those datasets need a re-ingest regardless — their stored targets were buckets and are unrecoverable.
- **Two tests pinned the old behavior** (`test_process_record_applies_bucket_label_policy`, and the `_map_unique_id` bucket assertions in `test_label_policy.py`). They now pin the boundary instead: the stored row keeps the raw value, the payload carries the buckets. No coverage was dropped.
- The three regression-class template READMEs gained one clause each, spelling out that bucketing is payload-only.
### Review round (saadqbal)

- **`data_samples[].label` type.** `apply_to_samples` returned the raw int bucket, making it a JSON number for bucketed datasets; pre-#486 it was the bucket read back out of the VARCHAR column, i.e. a string. The backend's summary serializer takes `samples` as an untyped `ListField` and would accept the number, but nothing downstream of `UserDataSet.data_samples` has ever seen one — so both helpers now emit stringified bucket ids and the wire shape is unchanged.
- **`get_label_counts` over raw continuous targets.** Attempted as a streamed fold, then reverted — the ceiling is documented at the call site instead, and #488 stays open. `mysql+mysqlconnector`'s SQLAlchemy dialect reports `supports_server_side_cursors = False`, so `stream_results` / `yield_per` are no-ops and the driver buffers the whole `GROUP BY` result regardless; chunking delivery via `Result.partitions()` bounded only the dict this code builds, not the driver's buffer (Bugbot caught that). The attempt also made `labels` polymorphic on the wire, which broke `e2e/test_characterization.py`'s `summary_kw["labels"].values()` — a fair warning about any observer of that call consuming the iterator before the real send. #488 carries the two shapes that would genuinely bound it: bucketing in SQL (`SHA2` + `CAST(CONV(...) AS UNSIGNED) % 64`, pinned against `_bucket`) or a DBAPI with server-side cursors.

- `TimeToEventValidator` still doesn't check that the event column is binary — worth a preflight guard, filed as a follow-up on #486 rather than widened into this PR.

## Verification

```
ruff check .                    All checks passed!
pytest tests/ -q                1955 passed, 1 xfailed
pytest tests/ --cov ... 95      Total coverage: 96.21% (gate: 95%)
```

New tests: bucket/passthrough behavior of both helpers (collision summing, row-total preservation, no input mutation, missing-label sentinel, unknown-policy raise); the payload shape out of `send_ingest_summary` for both policies; that an unknown policy raises before any POST; and an end-to-end `CSVIngestor` run of a `time_to_event_prediction` CSV under `bucket` asserting `0`/`1` reaches `insert_batch` while the summary call carries the policy — the reported failure, as a test.

---

> [!NOTE]
> **High Risk**
> Changes how regression-class labels are stored and how outbound ingest summaries are shaped — data integrity and privacy-boundary logic on a critical path. Existing regression datasets with bucketed stored targets remain broken until re-ingested.
> 
> **Overview**
> **Fixes #486:** `label.policy: bucket` was applied in `RecordProcessor` before INSERT, so regression-class datasets stored hash-bucket IDs instead of raw targets. Training then failed hard for time-to-event (event column must be 0/1) and silently trained on garbage for tabular regression / forecasting.
> 
> Bucketing now runs only in `APIClient.send_ingest_summary` via new `apply_to_label_counts` / `apply_to_samples` helpers. `RecordProcessor` no longer takes a policy and persists the raw label (with NaN/`pd.NA` normalized to SQL NULL). Classification `passthrough` behavior is unchanged.
> 
> Also bumps to **0.8.7** and clarifies in the three regression-class template READMEs that bucketing is payload-only. Note: content-hash `data_id`s change for regression sources because the stored record now includes the raw label — those datasets need a re-ingest anyway.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 292e93d198223f48c2b957f4f3f7c7218d6fe355. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/api/client.py
+++ b/tracebloc_ingestor/api/client.py
@@ -14,6 +14,7 @@ from ..utils.constants import (
     RED,
     YELLOW,
 )
+from ..utils import label_policy as label_policy_module
 
 # Logger for this module. Level is set by `setup_logging()` on the root
 # logger when the user script calls it; child loggers inherit that level.
@@ -225,6 +226,7 @@ class APIClient:
         samples: List[Dict[str, Any]],
         meta_data: Optional[Dict[str, Any]] = None,
         physical_table: Optional[str] = None,
+        label_policy: str = label_policy_module.PASSTHROUGH,
     ) -> Dict[str, Any]:
         """
         Send a single ingest summary to the backend, creating the UserDataSet in one
@@ -234,13 +236,22 @@ class APIClient:
         Args:
             table_name: Dataset table name (used as the URL path segment)
             ingestor_id: UUID identifying this ingest run
-            labels: ``{label: row_count}`` — computed locally after DB insert
+            labels: ``{label: row_count}`` — computed locally after DB insert,
+                keyed by the RAW label as stored
             dataset_title: Human-readable name for the new dataset
             data_format: One of "image", "tabular", "text", "audio", "video"
             data_intent: "train" or "test"
             category: TaskCategory value, e.g. "image_classification"
             schema: Column schema written to GlobalMetaData
             samples: Small list of representative records shown in the UI
+            label_policy: ``"passthrough"`` (classification — raw label values
+                ride the payload) or ``"bucket"`` (regression-class — every
+                outbound label becomes a stable hash-bucket id so the central
+                backend never sees the raw target). This is the ONLY place the
+                policy is applied: it is a boundary control, and the row in the
+                cluster's own MySQL keeps the raw target the training client
+                needs (#486). Both label-bearing payload fields — ``labels``
+                and ``samples`` — go through it.
             physical_table: RFC-0003 D16 (tracebloc/backend#1205) — the
                 per-ingestion physical table (``ds_<uuid4().hex>``, derived
                 from ingestor_id) this run wrote into, reported so the
@@ -254,7 +265,14 @@ class APIClient:
 
         Raises:
             requests.exceptions.RequestException: If the API call fails after retries
+            ValueError: If ``label_policy`` is not a known policy name.
         """
+        # The boundary. Applied before the local-mock branch too, so a dry run
+        # reports the label count that would actually ship (bucket collisions
+        # merge keys, so it can be lower than the raw count).
+        labels = label_policy_module.apply_to_label_counts(labels, label_policy)
+        samples = label_policy_module.apply_to_samples(samples, label_policy)
+
         if self.config.EDGE_ENV == "local":
             logger.info(
                 f"Mock: Would send ingest summary for {table_name} "
--- a/tracebloc_ingestor/api/client.py
+++ b/tracebloc_ingestor/api/client.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple, Union
+from typing import List, Dict, Any, Optional
 import os
 import requests
 import json
@@ -213,7 +213,7 @@ class APIClient:
         self,
         table_name: str,
         ingestor_id: str,
-        labels: Union[Mapping[str, int], Iterable[Tuple[Any, int]]],
+        labels: Dict[str, int],
         dataset_title: str,
         data_format: str,
         data_intent: str,
@@ -232,12 +232,8 @@ class APIClient:
         Args:
             table_name: Dataset table name (used as the URL path segment)
             ingestor_id: UUID identifying this ingest run
-            labels: Per-label row counts computed locally after DB insert, keyed
-                by the RAW label as stored. A mapping, or an iterable of
-                ``(label, count)`` pairs for callers that stream them
-                (``Database.iter_label_counts``) — either way it is folded once,
-                here, before any HTTP call, so a retry re-sends the built
-                payload rather than re-consuming the source
+            labels: ``{label: row_count}`` — computed locally after DB insert,
+                keyed by the RAW label as stored
             dataset_title: Human-readable name for the new dataset
             data_format: One of "image", "tabular", "text", "audio", "video"
             data_intent: "train" or "test"
--- a/tracebloc_ingestor/api/client.py
+++ b/tracebloc_ingestor/api/client.py
@@ -155,9 +155,7 @@ class APIClient:
             new = self.config.BACKEND_TOKEN or os.environ.get("BACKEND_TOKEN")
             if new and new != old:
                 self.token = new
-                logger.info(
-                    f"{GREEN}Re-read rotated BACKEND_TOKEN after 401.{RESET}"
-                )
+                logger.info(f"{GREEN}Re-read rotated BACKEND_TOKEN after 401.{RESET}")
                 return True
             return False
         # CLIENT_ID/PASSWORD path: mint a new token.
@@ -165,9 +163,7 @@ class APIClient:
             self.token = self.authenticate()
             return self.token != old
         except Exception as exc:
-            logger.error(
-                f"{RED}Failed to re-authenticate after 401: {exc}{RESET}"
-            )
+            logger.error(f"{RED}Failed to re-authenticate after 401: {exc}{RESET}")
             return False
 
     def _authed_request(
@@ -347,7 +343,9 @@ class APIClient:
                     f"HTTP {e.response.status_code}: {body}{RESET}"
                 )
             else:
-                logger.error(f"{RED}Error sending ingest summary: {str(e)[:500]}{RESET}")
+                logger.error(
+                    f"{RED}Error sending ingest summary: {str(e)[:500]}{RESET}"
+                )
             raise
 
     def get_dataset_metadata(self, ingestor_id: str) -> Optional[Dict[str, Any]]:
--- a/tracebloc_ingestor/api/client.py
+++ b/tracebloc_ingestor/api/client.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Any, Optional
+from typing import Any, Dict, Iterable, List, Mapping, Optional, Tuple, Union
 import os
 import requests
 import json
@@ -213,7 +213,7 @@ class APIClient:
         self,
         table_name: str,
         ingestor_id: str,
-        labels: Dict[str, int],
+        labels: Union[Mapping[str, int], Iterable[Tuple[Any, int]]],
         dataset_title: str,
         data_format: str,
         data_intent: str,
@@ -232,8 +232,12 @@ class APIClient:
         Args:
             table_name: Dataset table name (used as the URL path segment)
             ingestor_id: UUID identifying this ingest run
-            labels: ``{label: row_count}`` — computed locally after DB insert,
-                keyed by the RAW label as stored
+            labels: Per-label row counts computed locally after DB insert, keyed
+                by the RAW label as stored. A mapping, or an iterable of
+                ``(label, count)`` pairs for callers that stream them
+                (``Database.iter_label_counts``) — either way it is folded once,
+                here, before any HTTP call, so a retry re-sends the built
+                payload rather than re-consuming the source
             dataset_title: Human-readable name for the new dataset
             data_format: One of "image", "tabular", "text", "audio", "video"
             data_intent: "train" or "test"
--- a/tracebloc_ingestor/database.py
+++ b/tracebloc_ingestor/database.py
@@ -19,7 +19,6 @@ from sqlalchemy import (
     Index,
     bindparam,
     inspect,
-
 )
 from sqlalchemy.engine import Engine
 from sqlalchemy import LargeBinary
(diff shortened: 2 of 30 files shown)


tracebloc_ingestor/ingestors/record_processor.py
```diff
@@ -150,9 +154,7 @@ def _map_unique_id(
             # space-separated tags, etc.) pass through unchanged.
             if isinstance(label_val, str):
                 label_val = label_val.strip()
-            cleaned_record["label"] = label_policy_module.apply(
-                label_val, self.label_policy
-            )
+            cleaned_record["label"] = label_val
```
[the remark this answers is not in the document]
shujaatTracebloc: Real, and fixed in 12c22b7 — thanks.

You're right that dropping the write-path policy also dropped `_bucket`'s NaN absorption. The label branch only coerced numpy scalars and stripped strings, so an empty cell's float `nan` / `pd.NA` reached the binder for a VARCHAR(255) column.

Fix is the same missing-data normalization `process()` already applies to every schema column: null-like becomes SQL NULL (the column is `nullable=True`). Storing NULL rather than re-introducing a sentinel keeps the boundary semantics intact — `get_label_counts` reports a NULL label under the `""` key and `get_samples` as `""`, and both bucket to `MISSING_LABEL_BUCKET` in the payload. So a missing target still reaches the backend as the sentinel, and no sentinel is ever written into the on-prem row.

An empty-string label is left as-is on purpose: it stores fine and already buckets to the same sentinel, so classification behavior stays byte-identical.

Covered by five new tests: nan / `pd.NA` / `pd.NaT` normalize to NULL, `""` is preserved, `0` and `False` survive (guarding against a truthiness check), and the NULL → `""` → sentinel round-trip.

tracebloc_ingestor/api/client.py
```diff
@@ -254,7 +261,14 @@ def send_ingest_summary(
 
         Raises:
             requests.exceptions.RequestException: If the API call fails after retries
+            ValueError: If ``label_policy`` is not a known policy name.
         """
+        # The boundary. Applied before the local-mock branch too, so a dry run
+        # reports the label count that would actually ship (bucket collisions
+        # merge keys, so it can be lower than the raw count).
+        labels = label_policy_module.apply_to_label_counts(labels, label_policy)
```
saadqbal, line 269: Storing the raw target means `get_label_counts` now `GROUP BY`s over the raw label, so for a continuous regression target it materialises a dict with up to one entry per distinct value (~N rows) before this collapses it to ≤64. Pre-fix it grouped over the ≤64 stored buckets. Fine at typical sizes, but a large continuous-target dataset could blow up memory / the result set here — worth bucketing the counts in SQL, or at least noting the ceiling.

saadqbal (commented):
Really solid fix — the send boundary is the right home for this, and the tests pin it well (the end-to-end TTE repro especially). Two non-blocking things worth a look inline, plus a doc nit: the three regression READMEs now say the backend "never sees raw target values", which is true for the per-row labels but a touch overbroad — `feature_stats['label']` still ships the target's min/max/sum/count for regression-class tasks. Pre-existing, just more load-bearing now that this claim is spelled out.

tracebloc_ingestor/api/client.py
```diff
@@ -254,7 +261,14 @@ def send_ingest_summary(
...
+        samples = label_policy_module.apply_to_samples(samples, label_policy)
```
saadqbal, line 270: `apply_to_samples` returns int bucket ids, so `samples[].label` in the payload is now a JSON number for bucketed datasets — pre-fix it was the stored bucket read back as a VARCHAR string. If the backend treats `data_samples[].label` as a string (as it is for classification class names), worth confirming it tolerates the number.

shujaatTracebloc: Good catch — changed in d34e9b9 to keep it a string.

I checked the backend before deciding: `GlobalMetaSummarySerializer.samples` is an untyped `ListField` (`global_meta/serializers.py:149`), so it *would* accept the number, and `UserDataSet.data_samples` is a TextField holding `json.dumps` — no per-item validation anywhere. But every consumer downstream of it (`common/utils/dataset_utils.py` merge path, `UserDataSetSerializer.get_data_samples`, the preview UI) has only ever seen strings, because pre-#486 the bucket went into the VARCHAR `label` column and came back out of `get_samples` as one. Rather than bet on tolerance I matched the old wire shape exactly, so no consumer can tell the difference.

`apply_to_label_counts` keys are stringified for the same reason — JSON object keys are strings regardless, this just stops the in-process dict from disagreeing with what ships. Tests now assert `isinstance(label, str)` on the payload samples.

tracebloc_ingestor/api/client.py
shujaatTracebloc: Right, and the ceiling is real — before this PR the same query grouped over ≤ 64 stored buckets. Documented at the call site in d34e9b9 with the numbers (a 100k-row float target is a ~100k-entry dict, single-digit MB; a multi-million-row continuous target is where it matters), and filed as #488.

I deliberately didn't fix it here, because both shapes fight the invariant this PR exists to establish — the policy applied exactly once, at the boundary:

- **Folding at read time** (a key function into `get_label_counts`) puts the raw→bucket mapping back outside `send_ingest_summary`, and then the boundary has to be told to skip the counts but still bucket the samples. That is the same "who owns the policy" ambiguity that produced the original bug.
- **Bucketing in SQL** (`SHA2(label, 256)` + `CONV(..., 16, 10) % 64`) has to agree with Python's `_bucket` for every value or cross-run correlation breaks, and `CONV` saturates at unsigned 64-bit — a divergence there is worse than the memory cost.

#488 lists the shape I think actually works: let the boundary take an iterable of `(label, count)` and stream the query, so client memory is O(64) and the grouping stays server-side. It needs a new empty-result check in `BaseIngestor` (today `if not label_counts`), which is more than this fix should carry.

Happy to pull it into this PR instead if you'd rather not ship the ceiling — say the word.

shujaatTracebloc: Implemented rather than documented — 7cc01c4. The ceiling comment is gone; the counts now stream.

- `Database.iter_label_counts` yields `(label, count)` via `partitions()`, so the client holds one chunk (1000 rows) at a time instead of the whole ungrouped result. `get_label_counts` is now its materialised form — same query, same NULL normalisation, no duplicated SQL.
- `apply_to_label_counts` takes a mapping **or** an iterable of pairs and folds as it arrives, so peak memory is the ≤ 64 buckets. The policy is still applied exactly once, at the boundary — that was the constraint that made the other two shapes unattractive.
- `BaseIngestor` hands the boundary the stream. Two things that needed care: an iterator is always truthy, so the "inserted rows but no label counts" guard peeks one row and treats `None` as empty (same RuntimeError, now naming `iter_label_counts`); and the peek calls `iter()` first, otherwise a re-iterable — a list, a test double — would have its first row counted twice by the `chain`.

Server-side grouping is unchanged; that's MySQL's business. What's gone is the N-entry dict on the ingestor pod.

Test-side fallout worth flagging in review: mocked-Database tests that reach the summary stage were relying on an unstubbed MagicMock being truthy, so they now stub the pair form. Assertions are unchanged apart from naming the helper base.py actually calls. New tests cover the chunked fetch, the NULL normalisation, laziness, the pair/mapping equivalence, a 5000-value generator folding to ≤ 64 keys with no rows lost, and the pair form end-to-end through `send_ingest_summary`.

#488 is closed by this PR.

tracebloc_ingestor/database.py
```diff
@@ -976,35 +987,69 @@ def list_dataset_ingestor_ids(self) -> List[Dict[str, str]]:
...
                     f"FROM `{table_name}` "
                     f"WHERE ingestor_id = :ingestor_id "
                     f"GROUP BY label"
                 ),
                 {"ingestor_id": ingestor_id},
-            ).fetchall()
+            )
+            for chunk in result.partitions(self.LABEL_COUNT_FETCH_SIZE):
+                for label, cnt in chunk:
+                    yield (label if label is not None else "", cnt)
```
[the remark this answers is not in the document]
shujaatTracebloc: Correct, and it killed the approach — reverted in 4b5eb86.

You're right that `partitions()` chunks delivery, not the fetch. I went looking for the `yield_per` / `stream_results` fix you're implying and it doesn't exist for us: our DBAPI is `mysql+mysqlconnector`, and its SQLAlchemy dialect reports `supports_server_side_cursors = False` (verified against SQLAlchemy 2.0.50), so both options are no-ops and the driver buffers the whole GROUP BY result on execute. Chunked delivery would have bounded the dict I build while leaving the driver's buffer just as large — a memory bound in the docstring only, exactly as you describe.

The e2e suite then surfaced a second cost of the shape: passing an iterator as `labels` makes that payload field polymorphic and consumption-order-sensitive. `e2e/test_characterization.py` reads `summary_kw["labels"].values()`, and any observer of the call — a mock, a wrapper, a log line — either breaks on the iterator or consumes it out from under the real send.

So the ceiling is documented at the call site again, now with the driver fact that makes iterating a dead end, and #488 keeps the two shapes that would actually bound it: bucketing in SQL (`SHA2` + `CAST(CONV(...) AS UNSIGNED) % 64`, which has to agree with `_bucket` for every value, including the 64-bit boundary) or moving to a DBAPI with server-side cursors.

tracebloc_ingestor/api/client.py
shujaatTracebloc: Update, since I told you the opposite two commits ago: the streaming fix is reverted (4b5eb86) and the ceiling is documented again — with a much better reason than my first note.

Iterating cannot bound this on our stack. `mysql+mysqlconnector`'s SQLAlchemy dialect reports `supports_server_side_cursors = False` (2.0.50), so `stream_results` / `yield_per` are no-ops and the driver buffers the entire GROUP BY result on execute. My `partitions()` version chunked delivery only — it bounded the dict and left the driver's buffer identical, which Cursor Bugbot flagged as "memory bound in the docstring only". It was right.

The attempt also made `labels` polymorphic on the wire: e2e broke on `summary_kw["labels"].values()`, which is a fair warning about any observer of that call — a mock, a wrapper, a log line — either breaking on the iterator or consuming it before the real send.

#488 now carries the two shapes that would genuinely bound it:
1. Bucket in SQL — `SHA2(label, 256)` + `CAST(CONV(SUBSTR(...,1,16), 16, 10) AS UNSIGNED) % 64`, pinned by a test that asserts SQL and Python agree over a fixture value set including 64-bit boundary values. Note the trap: without the CAST, MySQL coerces `CONV`'s string through DOUBLE and loses the low bits.
2. Move to a DBAPI with server-side cursors (PyMySQL / MySQLdb) and stream for real.

Both are larger than this bug fix and neither is something I'd land without you weighing in on the DBAPI question — hence the ceiling plus #488 rather than a third attempt inside this PR.

saadqbal (approved):

PR #497, opened 2026-08-13
fix(validators): delete the three private column resolvers, route through the canonical rule (backend#1828)

LukasWodka:
Fixes tracebloc/backend#1828.

## The bug

`tracebloc_ingestor/utils/columns.py:resolve_column` is the one column-matching rule, reached by validators through `BaseValidator._match_column`. Three validators carried private `_resolve_column` copies instead, and **two dropped the `.strip()`**:

| copy | categories it gates |
|---|---|
| `BIOLabelValidator` | `token_classification` |
| `TabSeparatedRecordValidator` | `sentence_pair_classification`, `embeddings` |

`BaseValidator._load_data` reads the manifest with `pd.read_csv` and **no `skipinitialspace`**, so `filename, label` — the header Excel writes by default — parses as `["filename", " label"]`. The case-only copies never matched, so 3 of 16 categories rejected a dataset for a column visibly present in the file, while sibling validators in the *same preflight run* resolved it fine. This is the #340 / bugbot #252 divergence returning through a side door.

## Reproduced before the fix

```
raw parsed columns: ['filename', ' label']
BIOLabelValidator       valid=False errors=['Missing required column(s): label']
LabelDiversity(correct) valid=True  errors=[]      <- same header, sibling validator
```

`filename` is padded whenever it is not the first CSV field, which is what reaches the other two:

```
  parsed cols: ['label', ' filename']
  SentencePair  'label, filename'      valid=False errors=['Missing required column: filename']
  parsed cols: ['filename ', ' label']
  SentencePair  'filename , label'     valid=False errors=['Missing required column: filename']
  parsed cols: ['id', ' filename']
  Contrastive   'id, filename'         valid=False errors=['Missing required column: filename']
```

## How far the copies diverge

Differential run of each copy against canonical over a grid of header/name pairs:

```
LabelDiversity copy mismatches vs canonical: 0   <- faithful
BIO shadow      mismatches vs canonical: 23      <- the bug
```

A second divergence, measured: the shadows **raise** where canonical returns a hit.

```
canonical resolve_column(df.columns, '0') -> 0
shadow  _resolve_column(df, '0')          -> AttributeError: 'int' object has no attribute 'lower'
```

Not independently reachable through the CSV path (`pd.read_csv` always yields str headers), so I am **not** claiming it as a second live bug — but it is why patching the copies was the wrong fix.

## Design: delete, don't sync

All three copies are deleted and every call site routes through `self._match_column(df.columns, name)`.

The third copy (`LabelDiversityValidator._resolve_column`) was **faithful** — 0 divergences — and is deleted anyway. A faithful copy is exactly how the two broken ones came to exist; leaving one keeps the class of bug representable. Its `#252` rationale moved onto `_match_column`, which is now the documented sole resolver. (`_resolve_schema_column`, which maps over `self.schema` keys rather than a header, is a different lookup and is untouched.)

Two kinds of test, because the per-category regressions alone would not have caught this:

1. **Per affected category** — the padded-header case, exercised through the real CSV read path (a raw file, not a DataFrame), which is where the padding survives. The ticket notes the regression test existed only on the *one copy that was correct*, which is precisely how the divergence survived.
2. **`tests/test_no_shadowed_column_resolver.py`** — walks every `BaseValidator` subclass and fails if any defines a private resolver. A shadow is invisible to a reader (nothing about `self._resolve_column(...)` says "bypasses the canonical rule"), so the guard is mechanical, not a review convention. It includes a *guard-the-guard* assertion so a broken import cannot make it vacuously pass.

## Test evidence

Full suite, this branch:

```
4 failed, 1980 passed, 1 xfailed, 2 warnings in 19.75s
```

Baseline `origin/develop`, same machine:

```
4 failed, 1970 passed, 1 xfailed, 2 warnings in 20.66s
```

+10 tests, and **the same 4 failures** — all `float_overflow` cases, pre-existing and unrelated (a local numpy version difference; verified failing on a clean `origin/develop` worktree).

### Mutation testing — the fix is not inert

**1. Strip the `.strip()` from the canonical rule** (recreate the bug at its source):

```
FAILED tests/test_no_shadowed_column_resolver.py::test_match_column_delegates_to_the_canonical_rule
FAILED tests/test_bio_label_validator.py::test_csv_padded_header_still_resolves_label
FAILED tests/test_bio_label_validator.py::test_csv_padded_filename_header_still_resolves
FAILED tests/test_contrastive_pairs_validator.py::test_csv_padded_filename_header_still_resolves
FAILED tests/test_sentence_pair_validator.py::test_csv_padded_filename_header_still_resolves
FAILED tests/test_label_diversity_validator.py::test_csv_whitespace_header_still_checked
FAILED tests/test_label_diversity_validator.py::test_csv_whitespace_header_schema_na_rules_apply
FAILED tests/test_columns.py::test_resolve_column[cols2-label- label ]
FAILED tests/test_columns.py::test_resolve_column[cols4- label -label]
9 failed, 86 passed
```

**2. Re-shadow `_resolve_column` on `BIOLabelValidator`** (the exact regression the guard exists for):

```
FAILED tests/test_no_shadowed_column_resolver.py::test_no_validator_defines_a_private_column_resolver[_resolve_column]
FAILED tests/test_bio_label_validator.py::test_csv_padded_header_still_resolves_label
FAILED tests/test_bio_label_validator.py::test_csv_padded_filename_header_still_resolves
3 failed, 22 passed
```

Both the structural guard *and* the behavioral tests fire.

**3. Make class discovery return nothing** (would render the shadow checks vacuous):

```
FAILED tests/test_no_shadowed_column_resolver.py::test_validator_classes_are_discovered
E       assert 0 >= 20
1 failed, 5 passed
```

## Gates

- **Version bump**: `0.8.8` → **`0.8.9`**. Patch: a false-reject bug fix with no API or schema change; nothing added to the public surface.
- **`black` (26.3.1, diff-scoped)**: `9 files would be left unchanged`.
- **Heads-up on one unrelated hunk**: `tests/test_contrastive_pairs_validator.py` was **already** non-conforming on `origin/develop` (verified), and appending a test drags it into the diff-scoped gate. The reformat is one `write_text(...)` call collapsed onto a single line — noted so it doesn't read as scope creep.

## Not claimed

The ticket calls the two copies "byte-identical"; they differ in the annotation only (`df: pd.DataFrame` vs `df: Any`) — bodies are identical. The ticket also names the class `BioLabelValidator`; it is `BIOLabelValidator`. Neither affects the finding.

---

> [!NOTE]
> **Low Risk**
> Targeted preflight column-matching fix with no public API change; risk is limited to manifest header resolution behavior.
> 
> **Overview**
> Fixes **false “missing column” preflight failures** when CSV headers include spaces after commas (e.g. Excel’s `filename, label`), which `pd.read_csv` keeps as `" label"` / `" filename"` because `_load_data` does not use `skipinitialspace`.
> 
> **`BIOLabelValidator`**, **`TabSeparatedRecordValidator`** (sentence pairs / contrastive embeddings), and **`LabelDiversityValidator`** no longer use private `_resolve_column` helpers; they call **`BaseValidator._match_column`** → **`resolve_column`** (strip + case-insensitive match). The two broken copies only lower-cased and never matched padded headers, so **token_classification**, **sentence_pair_classification**, and **embeddings** could fail while other validators in the same run passed.
> 
> Adds **CSV regression tests** per affected validator and **`tests/test_no_shadowed_column_resolver.py`** to forbid new private column resolvers on any `BaseValidator` subclass. **`_match_column`** docs now state it is the sole resolver. Version **0.8.8 → 0.8.9**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26621113a0c905e7a20fded8eb2994004d4bcd2d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_bio_label_validator.py
+++ b/tests/test_bio_label_validator.py
@@ -185,3 +185,39 @@ def test_iob2_check_skipped_when_tag_format_invalid(validator, texts_dir):
     assert not result.is_valid
     assert any("invalid BIO tag" in e for e in result.errors)
     assert not result.warnings
+
+
+def test_csv_padded_header_still_resolves_label(texts_dir, tmp_path):
+    """``filename, label`` — the header Excel writes by default — must resolve,
+    not report the label column missing.
+
+    Regression (backend#1828): BIOLabelValidator carried a private
+    ``_resolve_column`` that lower-cased but did NOT strip, so pandas' parsed
+    header ``[" label"]`` never matched the configured ``label``. The dataset
+    was rejected for a column the user can see in their file, on
+    token_classification only, while sibling validators in the same preflight
+    run resolved it fine. Exercised through the real CSV read path (a raw file,
+    not a DataFrame) because that is where the padding survives —
+    ``BaseValidator._load_data`` calls ``pd.read_csv`` without
+    ``skipinitialspace``.
+    """
+    _write(texts_dir, "s1", "John Smith")
+    path = tmp_path / "padded.csv"
+    path.write_text("filename, label\ns1,B-PER I-PER\n", encoding="utf-8")
+
+    result = BIOLabelValidator().validate(str(path))
+
+    assert result.is_valid, result.errors
+    assert result.metadata["rows_checked"] == 1
+
+
+def test_csv_padded_filename_header_still_resolves(texts_dir, tmp_path):
+    """The same strip must apply to the ``filename`` column, which is padded
+    whenever it is not the first CSV field (backend#1828)."""
+    _write(texts_dir, "s1", "John Smith")
+    path = tmp_path / "padded_fn.csv"
+    path.write_text("label, filename\nB-PER I-PER,s1\n", encoding="utf-8")
+
+    result = BIOLabelValidator().validate(str(path))
+
+    assert result.is_valid, result.errors
--- a/tests/test_contrastive_pairs_validator.py
+++ b/tests/test_contrastive_pairs_validator.py
@@ -40,9 +40,7 @@ def _validate(filenames, staged):
 def test_valid_pair_and_triplet_pass(staged):
     texts, make_csv = staged
     (texts / "pair.txt").write_text("anchor text\tpositive text\n", encoding="utf-8")
-    (texts / "triplet.txt").write_text(
-        "anchor\tpositive\tnegative\n", encoding="utf-8"
-    )
+    (texts / "triplet.txt").write_text("anchor\tpositive\tnegative\n", encoding="utf-8")
     result = _validate(["pair", "triplet"], staged)
     assert result.is_valid, result.errors
     assert result.metadata["rows_checked"] == 2
@@ -186,3 +184,25 @@ def test_error_cap_summarizes(staged):
     assert not result.is_valid
     assert any("further errors suppressed" in e for e in result.errors)
     assert len(result.errors) <= 51
+
+
+def test_csv_padded_filename_header_still_resolves(staged, tmp_path):
+    """A padded ``filename`` header (``id, filename``) must resolve, not report
+    the column missing.
+
+    Regression (backend#1828): TabSeparatedRecordValidator — this validator's
+    base — carried a private ``_resolve_column`` that lower-cased but did NOT
+    strip, so pandas' parsed header ``" filename"`` never matched. ``embeddings``
+    then rejected a manifest for a column visibly present, while sibling
+    validators in the same preflight run resolved it. ``filename`` is padded
+    whenever it is not the first CSV field. Written as a raw file: pandas keeps
+    the space because ``_load_data`` does not pass ``skipinitialspace``.
+    """
+    texts, _ = staged
+    (texts / "pair.txt").write_text("anchor text\tpositive text\n", encoding="utf-8")
+    path = tmp_path / "padded.csv"
+    path.write_text("id, filename\n1,pair\n", encoding="utf-8")
+
+    result = ContrastivePairsValidator(texts_path="texts").validate(str(path))
+
+    assert result.is_valid, result.errors
--- a/tests/test_no_shadowed_column_resolver.py
+++ b/tests/test_no_shadowed_column_resolver.py
@@ -1,0 +1,104 @@
+"""Structural guard: no validator may carry its own column resolver.
+
+``tracebloc_ingestor.utils.columns.resolve_column`` is the single column-name
+matching rule, reached by validators through
+``BaseValidator._match_column``. Three validators once carried private
+``_resolve_column`` copies instead, and two of them dropped the ``.strip()`` —
+so ``token_classification`` / ``sentence_pair_classification`` / ``embeddings``
+false-rejected the header Excel writes by default (``filename, label``) as a
+missing column, while sibling validators in the SAME preflight run resolved it
+(backend#1828, reintroducing #340 / bugbot #252).
+
+The per-category regressions live next to each validator's own tests. This file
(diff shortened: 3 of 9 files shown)


shujaatTracebloc (approved):
Ran it rather than reading it. 86/86 green in a clean venv, and I mutation-tested the structural guard because a guard like this is worthless if it's vacuous:

```
baseline:                                        GREEN (no shadows present)
mutation 1 — faithful private copy re-added:     RED ✅
mutation 2 — .strip() dropped from canonical:    RED ✅
```

So both halves hold: `test_no_validator_defines_a_private_column_resolver` catches a *faithful* copy (which matters — the third copy, `LabelDiversityValidator`'s, was faithful and still deleted), and `test_match_column_delegates_to_the_canonical_rule` catches the divergence moving up one level into `resolve_column` itself. Deleting three copies without that second test would just have relocated the bug.

Verified the canonical rule in `utils/columns.py` genuinely subsumes what the copies did — exact match first, then stripped+lowered on both sides, returning the *original* spelling so callers can still index the raw frame. That last detail is what makes the `df` → `df.columns` signature change safe at all four call sites.

The `([0, 1], "0", 0)` case in the delegation test is a nice catch: the deleted copies did a bare `c.lower()` and would have raised `AttributeError` on a non-string column label, where the canonical rule returns a hit.

Two things I checked for risk and found clean:
- **Blast radius is one-directional.** This only makes resolution *more* permissive, so it can't reject a manifest that previously passed — it un-rejects the three categories that were false-negative. No back-compat concern.
- **`LabelDiversityValidator` behaviour is unchanged** — its copy already stripped, so only `token_classification` / `sentence_pair_classification` / `embeddings` actually change, which matches the PR's claim exactly.

Good framing on why the guard has to be mechanical: nothing at a `self._resolve_column(...)` call site tells a reader it bypasses the canonical rule, so review convention was never going to hold this. Version bump to 0.8.9 is there.

LGTM.

saqlainsyed007 (approved):
**APPROVE** — correctness review (focus: correctness).

## What this does
Deletes the three private `_resolve_column` copies on `BIOLabelValidator`, `TabSeparatedRecordValidator`, and `LabelDiversityValidator`, routing every call site through `BaseValidator._match_column` → `utils.columns.resolve_column` (the single strip + case-insensitive rule). Fixes false "missing column" preflight rejections on `token_classification` / `sentence_pair_classification` / `embeddings` for the padded headers `pd.read_csv` produces (e.g. `filename, label` → `" label"`), since two copies lower-cased without stripping.

## Correctness assessment — clean
- `resolve_column` is strictly more permissive than the two broken copies (adds `.strip()`) and identical to the faithful `LabelDiversityValidator` copy, so no behavior regresses and the three gated categories are fixed.
- Original (un-stripped) column name is returned, so callers still index the raw frame correctly (verified `BIOLabelValidator.validate` → `_validate_row`).
- Imports consistent: `Optional` dropped from `bio_label_validator.py` where now unused; retained in the other two where still referenced; `pd` still used (`pd.Series`).
- No private resolver remains in any validator; the new structural guard (`test_no_shadowed_column_resolver.py`) enforces this class of bug away and guards against a vacuous pass.

## Verification
Ran the affected suite in a clean venv on the PR head: `tests/test_no_shadowed_column_resolver.py`, the four validator suites, and `tests/test_columns.py` — **95 passed**. Matches the PR's mutation-testing evidence.

Tests, versioning (0.8.8 → 0.8.9, correct patch bump), and rationale are thorough. No changes requested.

PR #498, opened 2026-08-13
test(contract): pin the ingest→trainer storage contract (backend#1706)

shujaatTracebloc:
## Why

[tracebloc-engine#615](https://github.com/tracebloc/tracebloc-engine/issues/615): keypoint + semantic-segmentation training crashed on the first batch of **every CLI-ingested dataset**, because the trainer resolved images by `data_id` while this repo names the file after the manifest `filename`. The reader is fixed (engine#616). What was never fixed is *why the divergence survived two releases* — [backend#1706](https://github.com/tracebloc/backend/issues/1706).

Three facts have to stay in agreement, and nothing pinned any of them:

1. we write `DEST_PATH/<manifest filename><ext>`,
2. we store that stem in `filename` and something unrelated in `data_id`,
3. so every trainer dataset reader must key on `filename`.

When #350 flipped the default `data_id` strategy `uuid` → `content_hash`, nothing failed anywhere. It couldn't: the trainer's fixtures are hand-authored, and the shape they use (`data_id` set to the on-disk stem, **no `filename` column at all**) is one this repo has never produced. And the only two templates that set `unique_id_column="filename"` are `keypoint_detection` and `semantic_segmentation` — precisely the two categories whose readers keyed on `data_id`. That is not a coincidence; those templates were papering over the reader bug.

This PR is the **source-of-truth half**. tracebloc-engine carries the mirror and replays the capture through its dataset readers (paired PR linked below); the two are independent and can land in either order.

## What

| file | |
|---|---|
| `tracebloc_ingestor/storage_contract.py` | The framework column names, the `data_id` strategies, and the on-disk naming rule — one stdlib-only module the trainer can mirror verbatim, the same arrangement `identifiers.py` already uses for column-name legality. Wired into `image_transfer` and `RecordProcessor` so it is load-bearing, not documentation. |
| `tests/test_ingest_storage_contract.py` | Runs **real ingests** through the YAML/CLI route — real `resolve` → `_build_ingestor`, real validators, real record processing, real file transfer; only MySQL and the API are faked, each at its own boundary. 4 file-bearing CV categories × default + template-alias `data_id`. |
| `tests/fixtures/contracts/ingested_image_rows.json` | The capture those ingests produce. **Not a fixture** — regenerate with `UPDATE_INGEST_CONTRACT_GOLDEN=1` and carry it into tracebloc-engine `core/tests/contracts/`. |
| `e2e/test_image_lookup_contract_e2e.py` | The same contract through **real MySQL** — the trainer reads a table, not the dicts the unit suite captures, and the DDL/type/NULL/width gap in between is where a filename gets truncated. Resolves every SELECTed row with the trainer's own rule transcribed verbatim. Sibling of the semseg `mask_id` contract e2e (backend#816). |

### Behaviour change

None. Every literal replaced by a constant has an identical value, and the full suite is green either way. The version bump (0.8.8 → 0.8.9) is what `version-guard` requires for a `tracebloc_ingestor/` change.

### The assertions that matter

Both new suites carry a **negative control**, so neither can pass vacuously: under the default strategy `data_id` must name *no* file on disk. If that ever starts failing, something has re-aliased `data_id` to the on-disk name and every other assertion here has quietly gone hollow.

## Verification

```
pytest tests/     2021 passed, 1 xfailed
pytest e2e/         63 passed          (real MySQL, docker compose -f e2e/docker-compose.yml)
ruff check .        All checks passed
black --check       clean on every changed file
```

Refs tracebloc/backend#1706, tracebloc/tracebloc-engine#615

---

> [!NOTE]
> **Low Risk**
> No intentional runtime behavior change—constants replace identical literals; risk is mainly test/golden maintenance and keeping the mirrored contract in sync with tracebloc-engine.
> 
> **Overview**
> Pins the **ingest→trainer storage contract** so trainers resolve images via **`filename`**, not opaque **`data_id`** — the gap behind tracebloc-engine#615.
> 
> Adds **`storage_contract.py`** as the stdlib-only source of truth (column names, `data_id` strategies, on-disk naming helpers) for tracebloc-engine to mirror. **`file_transfer`** and **`RecordProcessor`** now use those constants instead of string literals; values are unchanged.
> 
> **`tests/test_ingest_storage_contract.py`** runs real ingests for four CV categories (default + `data_id` column alias), asserts naming/independence/semseg `mask_id`, and locks output in **`ingested_image_rows.json`** for cross-repo replay. **`e2e/test_image_lookup_contract_e2e.py`** repeats the check through real MySQL with the trainer’s `resolve_image_path` rule transcribed locally, including negative controls so `data_id`-only lookup cannot pass vacuously.
> 
> Docs note the new module and e2e contract suite; version **0.8.8 → 0.8.9**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91de0c8ed646fedbc47fa14ab911ef2def13f478. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,7 @@ the file extension — drives validation and file transfer.
   - `api/` -- API client for communicating with tracebloc backend
   - `database.py` -- MySQL/SQLAlchemy database operations
   - `file_transfer.py` -- secure file transfer to cluster storage
+  - `storage_contract.py` -- the ingest→trainer storage contract (framework column names, `data_id` strategies, the on-disk naming rule). Stdlib-only and mirrored verbatim by tracebloc-engine's `core/utils/ingest_contract.py`; change it here first (backend#1706)
   - `config.py` -- configuration
   - `utils/` -- shared utilities (incl. `TaskCategory` constants, logging)
 - **`templates/`** -- ingestor scripts used inside Docker containers
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -23,6 +23,21 @@ The suite **auto-skips when no MySQL is reachable**, so the default `pytest`
 (unit) run is unaffected. CI runs it with a MySQL service in
 `.github/workflows/e2e.yml`.
 
+## Cross-repo contract suites
+
+Two files here assert not just "the ingest succeeded" but "what it stored is
+what the **training client** resolves", checked against the client's own
+derivation rule (transcribed in each file, since neither repo may import the
+other):
+
+- `test_semseg_client_contract_e2e.py` — the semseg `mask_id` contract
+  (backend#816).
+- `test_image_lookup_contract_e2e.py` — the image-lookup contract for every
+  file-bearing CV category: the file is named after `filename`, and `data_id`
+  names nothing on disk (tracebloc-engine#615, backend#1706). Its unit-level
+  sibling is `tests/test_ingest_storage_contract.py`, which also produces the
+  capture tracebloc-engine replays through its dataset readers.
+
 ## Known gaps (currently `xfail`)
 
 None — all 11 supported modalities ingest cleanly and are covered. The earlier
--- a/e2e/test_image_lookup_contract_e2e.py
+++ b/e2e/test_image_lookup_contract_e2e.py
@@ -1,0 +1,272 @@
+"""backend#1706 — the ingestor→trainer IMAGE-lookup contract, end to end.
+
+Sibling of ``test_semseg_client_contract_e2e.py``, which pins the same kind of
+boundary for ``mask_id``. Here the subject is the image itself, and the bug it
+guards against already shipped: tracebloc-engine#615, where keypoint and
+semantic-segmentation training crashed on the first batch of every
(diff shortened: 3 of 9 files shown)


saadqbal (approved):
Really careful contract PR. The producer diff is pure literal→constant with identical values (`filename`/`extension`/`data_id`), and the 117 existing file_transfer/record_processor tests stay green — no behaviour change, as advertised. `storage_contract.py` faithfully encodes the real rule: `strip_image_extension` strips only a *recognised* suffix via the same gate as `file_transfer._has_extension` (not `split('.')[0]`), and `test_contract_agrees_with_the_transfer_code_it_describes` pins that against the actual transfer helper. It also lines up byte-for-byte with the engine mirror in tracebloc-engine#636 — same constants, same helper bodies, same golden hashes (`815352a9…`, `f2feb4ca…`) — so the two contract tests can't drift apart silently. Negative controls on both suites keep them from passing vacuously. Nice 👍

PR #502, opened 2026-08-14
ci(1606): tests.yml calls `make coverage`, and the "exactly" claim becomes true

LukasWodka:
Part of backend#1606, under epic backend#1680. Fourth repo in the CI-parity fan-out (backend#1962, client-runtime#329, averaging-service#346) — and the first where the two sides **genuinely differed** rather than being identical copies.

## A comment that asserted parity over a pair that had drifted

The Makefile said, in so many words:

```make
# coverage: tests.yml exactly — the 95% floor included.
```

It was not exactly:

| | Makefile `coverage` | `tests.yml` |
|---|---|---|
| scope | `pytest tests/` | **bare `pytest`** (also collects `e2e/`) |
| html report | — | `--cov-report=html` |

A comment claiming parity, over a pair that had drifted in both scope and output, is the *docstring-that-should-be-a-machine-check* pattern the workspace rules call out. This makes the claim structural: tests.yml calls the target.

## The scope moves CI-ward, not Makefile-ward — deliberately

`tests/` is a **safety decision** (Bugbot, #461). A bare `pytest` also collects `e2e/`, so on a developer machine with MySQL up — or `DB_NAME`/`DB_USER` exported — `make check` would quietly run real ingestion that **creates and drops tables**. Widening the Makefile to match CI would reintroduce exactly that hazard.

**Narrowing CI loses nothing measurable:**

- `e2e/conftest.py` skips the suite unless a MySQL is reachable, and the `pytest` job has no MySQL service — so those 6 files are collected and never executed
- the e2e suite has its **own required check**, `e2e (real MySQL)` (e2e.yml), which runs it against a real database

So this job stops collecting six files it was never going to run, and nothing stops being tested.

`--cov-report=html` moves into the target so the artifact upload still finds `htmlcov/`.

## Job name and matrix untouched

`pytest (3.11)` and `pytest (3.12)` are **required status checks**, matched by name. The job id, name and 3.11/3.12 matrix all re-parse unchanged.

## Verification

- `make -n coverage` now carries every flag this step had, including `--cov-report=html`
- 6 test files confirmed under `e2e/`, 91 under `tests/`; `e2e/conftest.py`'s skip-without-MySQL logic read directly
- YAML re-parsed: job name and matrix preserved

---

> [!NOTE]
> **Low Risk**
> CI-only test orchestration change; behavior tightens to the existing Makefile target with no production code touched.
> 
> **Overview**
> **CI now runs `make coverage` instead of duplicating pytest flags in the workflow**, so local `make check-all` and the `pytest` matrix job share one definition for the 95% floor, XML/term/HTML coverage output, and `tests/`-only collection.
> 
> The workflow step documents why scope narrows from bare `pytest` to **`tests/`**: e2e was only collected in CI without MySQL (skipped anyway) while a bare invoke on a dev machine with MySQL could run destructive ingestion (#461). **`make coverage` adds `--cov-report=html`** so the existing HTML artifact upload still works.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9808170cc7ac24e3b29722dfc7df9ad7a556be03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,13 +47,21 @@ jobs:
           pip install -e .
 
       - name: Run pytest with coverage
-        run: |
-          pytest \
-            --cov=tracebloc_ingestor \
-            --cov-report=term \
-            --cov-report=xml \
-            --cov-report=html \
-            --cov-fail-under=95
+        # `make coverage` (backend#1606). The Makefile's comment on that target
+        # claimed it was "tests.yml exactly"; it was not — this step invoked a
+        # bare `pytest` while the target scoped to tests/. Now the claim is
+        # structural instead of maintained by hand.
+        #
+        # That scope change applies here: a bare `pytest` also collected e2e/,
+        # which SKIPS ITSELF in this job for want of a MySQL service. The e2e
+        # suite has its own required check (`e2e (real MySQL)`) that runs it
+        # against a real database, so nothing stops being tested — this job just
+        # stops collecting six files it was never going to execute.
+        #
+        # The scope is not a detail to "fix" back: a bare pytest on a developer
+        # machine with MySQL up runs real ingestion that creates and drops
+        # tables (Bugbot, #461), which is why the Makefile is the narrow one.
+        run: make coverage
 
       - name: Coverage summary
         if: always()
--- a/Makefile
+++ b/Makefile
@@ -98,10 +98,28 @@ lint:
 test:
 	$(PYTEST) tests/ -q
 
-# coverage: tests.yml exactly — the 95% floor included.
+# coverage: tests.yml's pytest job exactly — the 95% floor and the HTML report
+# it uploads as an artifact.
+#
+# "Exactly" is now STRUCTURAL: tests.yml calls this target (backend#1606). It
+# claimed "exactly" before and was not — the workflow invoked a bare `pytest`
+# and added --cov-report=html, so the two differed in both scope and output
+# while a comment here asserted they did not.
+#
+# THE tests/ SCOPE NOW APPLIES TO CI TOO, and that is the right direction of
+# travel rather than a concession. Scoping is the deliberate decision from
+# `test` above (Bugbot, #461): a bare `pytest` also collects e2e/, which on a
+# developer machine with a MySQL up would run real ingestion that creates and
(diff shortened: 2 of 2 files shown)


shujaatTracebloc (approved):
Approve. backend#1606 parity, and it makes the Makefile's "tests.yml exactly" comment structurally true — CI called a bare `pytest` while the target scoped to tests/, differing in both scope and output. Now `make coverage`, with the tests/ scope applied to CI (correct direction: a bare pytest also collects e2e/, which self-skips here for lack of MySQL and has its own required `e2e (real MySQL)` check, and on a dev box with MySQL up would run real table-dropping ingestion — Bugbot #461). Makefile target gains the --cov-report=html CI was adding. Nothing stops being tested; six never-executed files stop being collected. CI green (incl. e2e + version-bump gate), Bugbot clean.

saqlainsyed007 (approved):
**APPROVE** — CI-only change, correct and well-reasoned.

**What it does:** `tests.yml`'s pytest job now runs `make coverage` instead of an inline pytest invocation, and the `coverage` Makefile target gains `--cov-report=html` so the two are structurally identical. This narrows CI collection from bare `pytest` (which also walked `e2e/`) to `pytest tests/`.

**Correctness verified:**
- **Coverage-neutral.** `e2e/conftest.py` uses `collect_ignore_glob = [...] if not _mysql_reachable() else []`, so with no MySQL service the e2e `test_*.py` files were never imported in this job — they contributed nothing to coverage before, and nothing is lost by no longer collecting them. The ~99% suite stays well above the unchanged `--cov-fail-under=95` floor.
- **e2e still covered.** Confirmed `e2e.yml` runs job `e2e (real MySQL)` against a real MySQL service — the separate check the safety argument relies on genuinely exists.
- **Artifacts intact.** `make coverage` runs from the repo root, so `.coverage` (Coverage summary step) and `htmlcov/` (upload-artifact step) land where the workflow expects them; adding `--cov-report=html` to the target is what keeps the upload working.
- **Required checks preserved.** Job name `pytest` and the 3.11/3.12 matrix are untouched, so `pytest (3.11)` / `pytest (3.12)` still match.

No correctness, convention, or coverage-regression issues found.

PR #503, opened 2026-08-14
chore(devex): install a `make check` pre-push hook via `make setup` (backend#1606)

LukasWodka:
Part of backend#1606, under epic backend#1680. Takes the pre-push hook fan-out from **10/16 repos to 16/16**.

## What this adds

`make install-hooks` — installs a git pre-push hook that runs `make check`, so "run the tests before you push" is carried by tooling rather than memory. `make setup` now calls it, and it stays independently runnable for a contributor who wants only the hook.

The target is **byte-identical to the one already in `backend`, `client-runtime`, `cli`, `model-zoo`, `tracebloc-engine` and `client`** — extracted verbatim rather than retyped, so this repo cannot start out drifted from the fleet.

## Honest by design

- Catches **forgetting, not defiance** — `git push --no-verify` skips it and always will.
- **Refuses to clobber** a pre-push hook that is already there and not ours (e.g. one `pre-commit` manages).
- **Skips delete/no-op pushes** — a branch delete streams an all-zeros local sha, so a red tree can't block `git push --delete`.
- **Degrades when `make` is absent** — GUI clients (Tower, GitKraken, VS Code) launch hooks with a minimal PATH and several don't expose `--no-verify`; skipping beats hard-blocking every push.
- Uses `git rev-parse --git-path hooks`, not a hard-coded `.git/hooks`, so it lands correctly in a linked worktree or submodule.
- Clears `GIT_DIR`/`GIT_WORK_TREE`/etc so a nested git call inside `make check` doesn't fail with status 128.

## Verification

Ran `make install-hooks` in a real checkout of this repo and asserted the hook is installed, executable, carries our marker, and its last line is `exec make check`.

Two refusal paths mutation-tested:

- planted a foreign `pre-push` and re-ran the target → **preserved**, with the "not ours — leaving it untouched" notice
- fed the hook an all-zeros local sha (a delete push) → **exit 0**, so cleanup pushes stay free

No CI behaviour changes: this adds a target and wires `setup` to it.

---

> [!NOTE]
> **Low Risk**
> Makefile-only developer workflow change; no runtime, CI, or application code paths are modified.
> 
> **Overview**
> Adds **`make install-hooks`**, which installs a git **pre-push** hook that runs **`make check`** so local pushes are linted and unit-tested before they leave the machine. **`make setup`** now invokes it automatically, and **`make help`** documents the new target.
> 
> The installer is defensive: it no-ops outside a git checkout, skips when **`core.hooksPath`** points at a shared hooks dir outside the repo, and **does not overwrite** an existing foreign pre-push hook. The generated hook skips delete/no-op pushes (all-zero SHAs), exits cleanly if **`make`** is missing (common with GUI git clients), and unsets **`GIT_*`** env vars before **`exec make check`** so nested git calls work in linked worktrees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ae4ec13fce9d2a283fd383f3ba8db78c9c08cb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ help:
 	@echo "  check       lint + the unit suite (~11 s) — run this before every push"
 	@echo "  check-all   everything CI runs, including the 95% coverage gate"
 	@echo "  setup       pip install -r requirements-dev.txt && pip install -e ."
+	@echo "  install-hooks  (re)install the git pre-push hook that runs 'make check'"
 	@echo
 	@echo "  individual: lint test coverage e2e"
 	@echo
@@ -69,6 +70,7 @@ setup:
 	$(PYTHON) -m pip install -e .
 	$(PYTHON) -m pip install "ruff==$(RUFF_VERSION)"
 	@echo "==> setup: dependencies installed; run 'make check'"
+	@$(MAKE) --no-print-directory install-hooks
 
 # ---- individual targets ------------------------------------------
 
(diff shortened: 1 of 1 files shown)


shujaatTracebloc (approved):
Approve (backend#1606). `make setup` now installs a pre-push hook running `make check`, factored into an independently-runnable `install-hooks`. Careful: uses `git rev-parse --git-path hooks` (worktree/submodule-safe), skips non-git checkouts, respects a `core.hooksPath` pointing outside the repo, and refuses to clobber a foreign pre-push hook. Honest that `--no-verify` bypasses it. CI green.

PR #513, opened 2026-08-19
sec(database): quote DB_NAME in CREATE DATABASE DDL (backend#952)

aptracebloc:
## What

Close the `DB_NAME` injection surface in `database.py`, on both halves of it:

1. **The DDL** — `CREATE DATABASE IF NOT EXISTS …` can't be parameterized, so `DB_NAME` is validated and backtick-quoted (doubling embedded backticks).
2. **The connection URL** — built with `URL.create()` / `URL.set(database=…)` instead of string concatenation, so SQLAlchemy escapes every component for its own grammar.

`DB_NAME` gets its **own** identifier grammar (`quote_database_identifier` in `identifiers.py`) rather than borrowing the column one.

## Why

`backend#952` — the `DB_NAME` half. The custom-processor ACE trust model, the other half of that issue, was split to **#2180** (a design call gated on `client#86`), so this PR closes #952.

The DDL was the filed finding, and on its own it's low severity — `DB_NAME` is operator/env-supplied, not user-facing. **Review found the more serious half:** `DB_NAME` was also interpolated raw into the connection URL, where a `?` is a query separator. Reproduced against live MySQL 8.0.46 with both `x` and `x?ssl_disabled=true` existing as real databases:

| path | `SELECT DATABASE()` | `Ssl_cipher` |
|---|---|---|
| before | `'x'` | `''` ← **TLS off** |
| after | `'x?ssl_disabled=true'` | `'TLS_AES_128_GCM_SHA256'` |

So the old path silently connected to a different database than the DDL had just created **and** handed `ssl_disabled=true` to the driver as a genuine connect arg — an env-driven TLS downgrade, neither failure raising.

Percent-encoding the path segment is not sufficient: `make_url` decodes userinfo but leaves the path escaped, so the driver receives the literal `%3F` name — a different database again. Measured on the same server, the encoded form raises `1049 Unknown database` for names that legitimately exist (`` a`b ``, `db name`). Components-in is the only form that round-trips.

Removing the hand-rolled `quote()` also fixes an asymmetry the review flagged in passing: `DB_USER` was interpolated raw while `DB_PASSWORD` beside it was quoted, so a `:` or `@` in the username mangled the userinfo section.

## Why a separate database grammar

Reusing `quote_column_identifier` borrowed a column-specific contract:

- the operator-facing error said `Column name '…' exceeding the 64-character database limit` for a `DB_NAME` misconfiguration — and since this validates before connecting, that message is their only diagnostic;
- it coupled `DB_NAME` legality to a grammar reconciled with `tracebloc-engine` and pinned to `schema/column_identifier.v1.json` (ISSUE #382), so loosening it for a customer CSV header would silently loosen `DB_NAME` too.

`MAX_DATABASE_IDENTIFIER_LENGTH` is therefore its own literal, quoting is shared via `_backtick_quote`, and the column entry points are behaviourally unchanged — the `#382` pin test and the trainer's verbatim mirror are unaffected.

**No character rejections were added, deliberately.** Tested against live MySQL **5.7.44 and 8.0.46**: both *accept* `/`, `\`, `.`, backtick, space and `?` in a backtick-quoted database name (filename-encoded on disk) and reject only length (`ERROR 1059` at 65). The two constraint sets are identical today; rejecting those characters would refuse names the server itself accepts. Recorded in the module comment and `tests/test_database_identifier.py` so it isn't re-derived.

## Tests

`make check-all` green — **2098 passed**, 1 xfailed, coverage **96.23%** (gate 95%).

- `tests/test_database.py` — **79 passed** (was 66). New: URL-shape assertions on both engines (server-level engine carries no database), a parametrised regression that `DB_NAME` can't smuggle connect args into the URL, `DB_USER`/`DB_PASSWORD` round-tripping through userinfo, and DB-worded validation errors.
- `tests/test_database_identifier.py` — new, covers the grammar directly, including the accept-list verified against real MySQL and the independence of the two length constants.
- Migration verified behaviour-preserving: driver kwargs are byte-identical for every valid config, including passwords with `@`, `:`, `/` and a pre-encoded `%40`.

## Notes

- The pre-existing whole-file black drift in `database.py` was adopted in a separate commit (CI black is diff-scoped); it's semantically neutral.
- `e2e (real MySQL)` exercises the new URL construction against MySQL 8.0 in CI.

Closes #952. Refs #2180.

### DIFF
--- a/tracebloc_ingestor/database.py
+++ b/tracebloc_ingestor/database.py
@@ -19,7 +19,6 @@ from sqlalchemy import (
     Index,
     bindparam,
     inspect,
-
 )
 from sqlalchemy.engine import Engine
 from sqlalchemy import LargeBinary
@@ -99,9 +98,7 @@ def _execute_with_retry(connection, stmt):
             # Swallow it so the original transient error propagates to
             # tenacity for the retry (or, on the last attempt, to the
             # caller's existing error path).
-            logger.debug(
-                f"connection.rollback() failed between retries: {rb_exc}"
-            )
+            logger.debug(f"connection.rollback() failed between retries: {rb_exc}")
         raise
 
 
@@ -148,7 +145,7 @@ class Database:
 
     def _get_sqlalchemy_type(self, mysql_type: str):
         """Convert MySQL type to SQLAlchemy type.
-        
+
         Extracts the base type (before parentheses) and matches exactly to avoid
         substring issues (e.g., "DATE" matching "DATETIME").
         """
@@ -175,10 +172,10 @@ class Database:
             "BLOB": BLOB,
             "LONGBLOB": LONGBLOB,
         }
-        
+
         mysql_type_upper = mysql_type.upper().strip()
         base_type = mysql_type_upper.split("(")[0].split()[0]
-        
+
         if base_type in type_mapping:
             alchemy_type = type_mapping[base_type]
             # Extract parenthesised arguments. Two shapes:
@@ -267,8 +264,16 @@ class Database:
         # is intentionally excluded — it's the user-facing label column the
         # framework maps onto the standard `label` column.
         _RESERVED = {
-            "id", "created_at", "updated_at", "status", "data_intent",
-            "data_id", "filename", "extension", "annotation", "ingestor_id",
+            "id",
+            "created_at",
+            "updated_at",
+            "status",
+            "data_intent",
+            "data_id",
+            "filename",
+            "extension",
+            "annotation",
+            "ingestor_id",
         }
         _collisions = sorted(_RESERVED & set(schema))
         if _collisions:
@@ -358,9 +363,17 @@ class Database:
             # to work around an unrelated error is exactly this case). Surface an
             # actionable error naming the drift instead.
             _STANDARD_COLUMNS = {
-                "id", "created_at", "updated_at", "status", "label",
-                "data_intent", "data_id", "filename", "extension",
-                "annotation", "ingestor_id",
+                "id",
+                "created_at",
+                "updated_at",
+                "status",
+                "label",
+                "data_intent",
+                "data_id",
+                "filename",
+                "extension",
+                "annotation",
+                "ingestor_id",
             }
             existing_features = {c.name for c in table.columns} - _STANDARD_COLUMNS
             expected_features = set(schema) - _STANDARD_COLUMNS
@@ -524,9 +537,7 @@ class Database:
                 # in #191 but dropped by the squash-merge — re-applying).
                 insert_stmt = insert(table)
                 update_dict = {
-                    column.name: text(
-                        f"VALUES(`{column.name.replace('`', '``')}`)"
-                    )
+                    column.name: text(f"VALUES(`{column.name.replace('`', '``')}`)")
                     for column in table.columns
                     if column.name not in ["id", "created_at", "data_id"]
                 }
@@ -584,9 +595,7 @@ class Database:
                             result["failures"].append(
                                 {
                                     "record": record,
-                                    "error": redaction.safe_db_error(
-                                        individual_error
-                                    ),
+                                    "error": redaction.safe_db_error(individual_error),
                                 }
                             )
                             connection.rollback()
@@ -970,7 +979,9 @@ class Database:
                             f"WHERE ingestor_id IS NOT NULL AND ingestor_id <> ''"
                         ),
                     ).fetchall()
-            except Exception as exc:  # noqa: BLE001 — skip the table, continue the sweep
+            except (
+                Exception
+            ) as exc:  # noqa: BLE001 — skip the table, continue the sweep
                 logger.warning(
                     "list_dataset_ingestor_ids: skipping table %s — could not read "
                     "ingestor_ids (%s)",
@@ -1180,7 +1191,10 @@ class Database:
                 ),
                 {"ingestor_id": ingestor_id, "limit": limit},
             ).fetchall()
-        return [{"data_id": row[0], "label": row[1] if row[1] is not None else ""} for row in rows]
+        return [
+            {"data_id": row[0], "label": row[1] if row[1] is not None else ""}
+            for row in rows
+        ]
 
     def get_table_schema(self, table_name: str) -> Dict[str, str]:
         """
@@ -1220,10 +1234,7 @@ class Database:
 
             # MySQL has no native BOOLEAN type: BOOL columns are created — and
             # therefore reflected — as TINYINT(1).
-            if (
-                type_name == "TINYINT"
-                and getattr(col_type, "display_width", None) == 1
-            ):
+            if type_name == "TINYINT" and getattr(col_type, "display_width", None) == 1:
                 schema[column["name"]] = "BOOLEAN"
                 continue
 
--- a/tracebloc_ingestor/database.py
+++ b/tracebloc_ingestor/database.py
@@ -20,7 +20,7 @@ from sqlalchemy import (
     bindparam,
     inspect,
 )
-from sqlalchemy.engine import Engine
+from sqlalchemy.engine import URL, Engine
 from sqlalchemy import LargeBinary
 from sqlalchemy.dialects.mysql import insert, LONGBLOB, BLOB
 from sqlalchemy.exc import OperationalError, InterfaceError, DBAPIError
@@ -28,7 +28,6 @@ import logging
 import secrets
 
 from .utils import redaction
-from urllib.parse import quote
 from typing import List, Dict, Any, Optional
 from datetime import datetime
 from tenacity import (
@@ -39,7 +38,7 @@ from tenacity import (
     before_sleep_log,
 )
 from .config import Config
-from .identifiers import MAX_COLUMN_IDENTIFIER_LENGTH, quote_column_identifier
+from .identifiers import MAX_COLUMN_IDENTIFIER_LENGTH, quote_database_identifier
 from .utils.typo_suggest import suggest_type as _suggest_type
 
 # Configure unified logging with config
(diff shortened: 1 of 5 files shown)


tracebloc_ingestor/database.py
```diff
@@ -129,8 +126,14 @@ def _create_engine(self) -> Engine:
...
+            # can't be parameterized; backtick-quote it via the shared
+            # MySQL-identifier quoter so it can't inject (backend#952).
             connection.execute(
-                text(f"CREATE DATABASE IF NOT EXISTS {self.config.DB_NAME}")
+                text(
+                    "CREATE DATABASE IF NOT EXISTS "
+                    f"{quote_column_identifier(self.config.DB_NAME)}"
+                )
             )
             connection.commit()
```
shujaatTracebloc, line 145: **`DB_NAME` is still interpolated raw into the connection URL — the injection surface is only half closed, and this is the half that fails silently.**

The DDL above is now safe, but line 141 (just below this hunk) is unchanged:

```python
connection_string = f"{base_connection_string}/{self.config.DB_NAME}"
```

Concrete scenario — operator/env sets `DB_NAME="training_test_datasets?ssl_disabled=true"`. The DDL side behaves as this PR intends: MySQL creates a database literally named that. Then SQLAlchemy parses the URL path, and the `?` is a query separator. Verified against the SQLAlchemy pinned in this repo:

```
>>> make_url("mysql+mysqlconnector://tb:pw@h:3306/training_test_datasets?ssl_disabled=true")
database = 'training_test_datasets'   query = {'ssl_disabled': 'true'}
```

Two failures at once, neither of which raises:

* the returned engine connects to a **different database** than the one just created — rows land in the wrong schema, or you get an "unknown database" much later at first table access, far from the cause;
* `ssl_disabled=true` is handed to `mysqlconnector` as a genuine connect arg — an **env-driven TLS downgrade** on the DB connection, which is squarely inside backend#952's threat model.

Any `?` in `DB_NAME` triggers this; `#` and `/` mangle the path in related ways (`make_url` keeps them in `database`, then the server rejects or misresolves).

Fix is the same one already applied to the password on line 118 — percent-encode the path segment:

```python
        # Now connect to the specific database
        connection_string = f"{base_connection_string}/{quote(self.config.DB_NAME, safe='')}"
```

Note this must be URL-encoding, not backtick-quoting — a URL path segment and a MySQL identifier are different grammars, so `quote_column_identifier` is the wrong tool here (backticks in the path would be sent to the server as part of the name).

While you're in this line: **`DB_USER` on line 118 is also raw** while `DB_PASSWORD` next to it is `quote()`d — a `:` or `@` in the username mangles the userinfo section the same way. Same one-line fix, worth doing in this pass so the URL construction is uniformly encoded.

tracebloc_ingestor/database.py
```diff
@@ -129,8 +126,14 @@ def _create_engine(self) -> Engine:
...
 
         with engine.connect() as connection:
+            # DB_NAME is operator/env-supplied and interpolated into DDL that
+            # can't be parameterized; backtick-quote it via the shared
+            # MySQL-identifier quoter so it can't inject (backend#952).
             connection.execute(
-                text(f"CREATE DATABASE IF NOT EXISTS {self.config.DB_NAME}")
+                text(
+                    "CREATE DATABASE IF NOT EXISTS "
+                    f"{quote_column_identifier(self.config.DB_NAME)}"
```
shujaatTracebloc: **Reusing `quote_column_identifier` borrows a column-specific contract, with two real consequences.**

**1. The operator-facing error names the wrong thing.** With a 70-character `DB_NAME`, `Database.__init__` now raises before connecting:

```
InvalidColumnIdentifierError: Column name 'training_…' is 70 characters,
exceeding the 64-character database limit — shorten it.
```

The operator misconfigured the `DB_NAME` env var; the message sends them hunting through their CSV headers instead. Same for the NUL and empty-string branches. This is a new failure mode (previously MySQL rejected it), so the message is the only diagnostic the operator gets.

**2. It couples `DB_NAME` validation to a grammar that is co-owned with another repo.** `identifiers.py` documents itself as the authoritative *column* grammar, deliberately mirrored verbatim by `tracebloc-engine` and pinned to `schema/column_identifier.v1.json` by `tests/test_column_identifier_schema.py` (ISSUE #382). After this change, any future loosening of that reconciled grammar silently loosens DB_NAME validation too, and any DB-name-motivated tightening breaks the trainer contract — the exact silent-divergence class #382 was closed to prevent. The grammars also genuinely differ today: MySQL rejects `/`, `\`, and `.` in **database** names even when backtick-quoted (they map to directory names, `ERROR 1102`), whereas this validator accepts all three.

Suggest a thin, correctly-named entry point rather than reaching into the column API — e.g. a `quote_database_identifier(name)` in `identifiers.py` that shares the quoting mechanic (`"`" + name.replace("`", "``") + "`"`) but raises a DB-worded error, or a neutral `quote_mysql_identifier(name, kind)` that both paths call. You keep the defense-in-depth this PR is after, without the misleading contract or the cross-repo coupling.

shujaatTracebloc (changes requested):
Reviewed for correctness. The quoting change itself is right (`quote_column_identifier` doubles embedded backticks correctly), the two new tests assert the DDL precisely, and I verified the second commit's black reformat is semantically neutral — the ASTs of `database.py` before/after are identical modulo docstring whitespace. `tests/test_database.py`: 66 passed locally on 6978790.

Two correctness findings, both about the DB_NAME identifier path:

1. **The injection surface is only half closed** — the DDL is quoted, but the connection URL three lines below still interpolates `DB_NAME` raw, and that half fails *silently* (wrong database + env-driven TLS downgrade).
2. **Borrowing the column-identifier contract** produces an operator-facing error that says "Column name …" for a `DB_NAME` misconfiguration, and couples DB_NAME validation to a grammar that is reconciled with tracebloc-engine and pinned by schema tests.

Details inline.

shujaatTracebloc:
@aptracebloc — friendly ping: this has been sitting at `CHANGES_REQUESTED` since my [review](https://github.com/tracebloc/data-ingestors/pull/513#pullrequestreview-4971441803) with no activity, and the head is still 6978790.

Two open threads, both on `database.py`:

1. **[line 141 area](https://github.com/tracebloc/data-ingestors/pull/513#discussion_r3812417389)** — `DB_NAME` is still interpolated raw into the SQLAlchemy connection URL, so the injection surface the PR set out to close is only half closed. A `?` in `DB_NAME` splits into query args: the engine connects to a *different* database than the DDL created, and something like `ssl_disabled=true` reaches mysqlconnector as a real connect arg. Fix is the `quote()` already applied to the password one line up (`DB_USER` needs it too).
2. **[line 135](https://github.com/tracebloc/data-ingestors/pull/513#discussion_r3812417400)** — reusing `quote_column_identifier` reports `DB_NAME` misconfigurations to operators as `Column name '…' is 70 characters…`, and couples DB_NAME validation to the #382 column grammar that's mirrored in tracebloc-engine and pinned by the schema test.

Happy to re-review as soon as there's a push, or to talk through (2) if you'd rather keep the shared helper and just reword — reply on the thread either way so it isn't blocking silently.

aptracebloc:
Thanks — finding 1 was a real bug and the more serious half of the two. Both are addressed in `ad9f6cd` + `dd18c78`. I implemented finding 1 differently from the suggestion, with evidence below, and one premise in finding 2 turned out not to hold.

## Finding 1 — confirmed, and worse than "wrong database"

Reproduced against **live MySQL 8.0.46** with both `x` and `x?ssl_disabled=true` existing as real databases, so the two outcomes are distinguishable:

| path | `SELECT DATABASE()` | `Ssl_cipher` |
|---|---|---|
| **OLD** (concatenated) | `'x'` | `''` ← **TLS off** |
| **NEW** (`URL.create`) | `'x?ssl_disabled=true'` | `'TLS_AES_128_GCM_SHA256'` |

So the TLS downgrade wasn't just theoretically reachable — the connection genuinely came up unencrypted. Good catch.

## Why not `quote(DB_NAME, safe="")`

The suggested fix closes the TLS vector but doesn't round-trip, because **`make_url` decodes userinfo but leaves the path segment escaped**. The driver then receives the literal percent-encoded string, which is a *different* database from the one the DDL just created — the same class of bug, relocated.

Measured on the same server, for names that legitimately exist there:

| `DB_NAME` | `quote(safe="")` | `URL.create` |
|---|---|---|
| `x?ssl_disabled=true` | `1049 Unknown database` | connects, TLS on |
| `` a`b `` | `1049 Unknown database` | connects |
| `db name` | `1049 Unknown database` | connects |

For the injection case failing closed is arguably fine, but it also refuses valid operator configs. Building the URL from components has SQLAlchemy escape each part for its own grammar and round-trips every case — so `_create_engine` now uses `URL.create(...)` and `base_url.set(database=...)`.

That also subsumes your aside: the hand-rolled `quote()` is gone entirely, so `DB_USER` is no longer raw next to a quoted `DB_PASSWORD`. I checked the migration is behaviour-preserving for valid configs — driver kwargs are byte-identical for passwords containing `@`, `:`, `/` and a pre-encoded `%40`.

## Finding 2 — substance taken, one premise corrected

Agreed on both consequences, and both are fixed: there's now `validate_database_identifier` / `quote_database_identifier` / `is_valid_database_identifier` with DB-worded errors ("shorten DB_NAME", never "Column name"), plus its own `MAX_DATABASE_IDENTIFIER_LENGTH` literal so the pinned column grammar can move without dragging DB_NAME. Quoting is shared via `_backtick_quote`, and the column entry points are behaviourally untouched — the `#382` pin test and the trainer's verbatim mirror are unaffected.

**The grammar-divergence premise doesn't hold, though.** The claim was that MySQL rejects `/`, `\` and `.` in database names even when backtick-quoted (`ERROR 1102`). Tested on **5.7.44 and 8.0.46**:

```
slash      'g_db/other'            -> CREATED
backslash  'g_db\other'            -> CREATED
dot        'g_db.other'            -> CREATED
backtick   'g_a`b'                 -> CREATED
space      'g_db name'             -> CREATED
question   'g_x?ssl_disabled=true' -> CREATED
len_64     (64 chars)              -> CREATED
len_65     (65 chars)              -> ProgrammingError 1059: Identifier name ... is too long
```

Both versions accept all of them — they're filename-encoded on disk — and both reject only length. So the database and column constraint sets are in fact identical today, and I deliberately did **not** add character rejections: they'd refuse names the server itself accepts, and `ERROR 1102` would have been a fix built on a false premise. The decoupling argument still stands on its own (independent constants, so the two can diverge later without silence), which is why it's implemented that way — the evidence is recorded in the module comment and `tests/test_database_identifier.py`'s docstring so nobody re-derives it.

## Verification

`make check-all` green — **2098 passed**, 1 xfailed, coverage **96.23%** (gate 95%). `tests/test_database.py` 79 passed, up from 66; new `tests/test_database_identifier.py` covers the grammar directly. New tests: URL-shape assertions on both engines, a parametrised connect-arg-smuggling regression, `DB_USER`/`DB_PASSWORD` round-tripping, and the DB-worded error messages.

Ready for another look.

— drafted with Claude Code

shujaatTracebloc (approved):

saadqbal (approved):

shujaatTracebloc (approved):
Re-reviewed `ad9f6cd`, `dd18c78`, `7bfc00c` independently. Both findings are genuinely fixed, and on the one point where we disagreed, **you're right and I was wrong** — details below.

### Finding 1 — fixed, and your approach is better than my suggestion

I verified the end state rather than the URL shape, by letting the real mysqlconnector dialect build the connect args for `DB_NAME="x?ssl_disabled=true"`, `DB_USER="tb@x"`, `DB_PASSWORD="p@ss:w/rd"`:

```
engine.url.database: 'x?ssl_disabled=true'
connect args: {'host': ..., 'database': 'x?ssl_disabled=true', 'user': 'tb@x',
               'password': 'p@ss:w/rd', 'port': 3306, ...}
```

Name verbatim, empty query, nothing smuggled, and user/password round-trip through the characters that used to mangle the userinfo. I also confirmed your objection to my suggested fix: `make_url` leaves the path segment escaped, so `quote(DB_NAME, safe="")` hands the driver `x%3Fssl_disabled%3Dtrue` — a different database from the one the DDL creates. Building from components is the right call; my suggestion would have traded a TLS bug for a wrong-database bug.

### Finding 2 — fixed, and my `/` `\` `.` premise was wrong

Separate constant, separate error type, DB-worded messages, shared `_backtick_quote`, column entry points untouched — that's the decoupling, and it doesn't disturb the #382 pin.

On the grammar-divergence claim: **retracting it.** Your measurement is consistent with the current manual, which I should have checked before asserting it. "Mapping of Identifiers to File Names" states plainly that any character is legal in a database identifier except ASCII NUL, and that problematic characters are hex-encoded into the directory name (`0x003F -> @003f`) rather than rejected. The `/`-and-`\`-are-rejected rule I was recalling predates that encoding. Deliberately adding no character rejection is correct, and the recorded evidence in the module comment is accurate as written — thanks for testing it instead of taking my word for it.

### Verification

Full suite on `7bfc00c`: **2098 passed, 1 xfailed** — matches your numbers. `black --check` clean on all four touched files. `MAX_COLUMN_IDENTIFIER_LENGTH` is still used at `database.py:306`, so the narrowed import isn't dead.

No further findings. Approving.

saadqbal:
/fr-pass

Journey-covered. Run [32356350119](https://github.com/tracebloc/e2e-test-agent/actions/runs/32356350119) is green against staging (09:55 -> 10:17 UTC, after the 09:11-09:13 merges), and the journey ingests a dataset end to end, so the `CREATE DATABASE` DDL path this PR quotes is exercised rather than reasoned about.

PR #514, opened 2026-08-19
sec(2157): scope the add-to-kanban App token to least privilege

LukasWodka:
Refs backend#2157. One of 18 — `add-to-kanban.yml` is a **byte-compared copy**, so least privilege cannot be applied to one repo without splitting the fleet.

## Why this exists

Without any `permission-*`, the mint carried the App's **full installation grant** — contents + pull-requests write across every installed repo. And the App holds `bypass_reviews` on staging and prod fleet-wide, so the blast radius was *merge past review*, not merely write.

```yaml
repositories: ${{ github.event.repository.name }}   # reads stop being org-wide
permission-issues: read                             # add-to-project must RESOLVE
permission-pull-requests: read                      #   the node before adding it
permission-organization-projects: write
permissions: {}                                     # the job needs no GITHUB_TOKEN
```

## Byte-identical to `backend`'s copy, deliberately

That copy is the one that survived review. @saadqbal caught that `owner:` narrows nothing — the run log says it verbatim, `Input 'repositories' is not set. Creating token for all repositories owned by tracebloc` — and @aptracebloc caught a run I had cited as evidence that was not one. Copying the corrected version instead of re-deriving it is the whole point of a byte-compared file.

## Verified, not assumed

The open question was whether `repositories:` scoping clips `organization_projects` (an org-level permission). It does not: [run 32255581084](https://github.com/tracebloc/backend/actions/runs/32255581084) on backend#2181's head exercised these exact scopes and **landed the card** — `Status=Code review`. A landed card, not a passing mint, is the assertion.

## The drift window

`caller-drift` compares copies against `.github`'s, so findings **rise** as this sweep lands and reach 0 only when `.github` goes last. That is deliberate and is the same ordering the App migration used. **Do not dispatch caller-drift remediation during the window** — it would propose reverting these files.

Measured before starting: the fleet was split three ways, and the audit was already red at 2 findings ([run 32268082510](https://github.com/tracebloc/.github/actions/runs/32268082510)) — `backend` and `frontend-app` carried *different* partial fixes. `frontend-app`'s was missing `repositories:` entirely, so it looked scoped and its reads were still org-wide.

---

> [!NOTE]
> **Medium Risk**
> Changes CI authentication for org project writes; behavior was validated on backend but mis-scoping would break kanban automation fleet-wide via the byte-identical workflow copy.
> 
> **Overview**
> **Hardens the engineer kanban workflow** so it no longer relies on broad credentials when issues or PRs open.
> 
> The job now sets **`permissions: {}`**, so the default `GITHUB_TOKEN` is unused; board actions use only the **tracebloc-release-train** App installation token.
> 
> That mint step is narrowed with **`repositories: ${{ github.event.repository.name }}`** (so reads are not org-wide) and explicit **`permission-issues` / `permission-pull-requests` read** plus **`permission-organization-projects` write**, matching what `actions/add-to-project` needs to resolve the triggering node and add it to the org project. Inline comments document prior failure modes and what was verified vs assumed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1076ef11ff76f45fc04f60a4b0a9dbcae2ccd541. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/add-to-kanban.yml
+++ b/.github/workflows/add-to-kanban.yml
@@ -9,6 +9,11 @@ on:
 jobs:
   add-to-project:
     runs-on: ubuntu-latest
+    # NO GITHUB_TOKEN AT ALL (saadqbal, #2181). Every call in this job authenticates
+    # as the App, so the workflow token needs nothing -- and an empty grant is the
+    # only version of that claim a reader can check. Free, and it means the least-
+    # privilege story covers both credentials in the job rather than just the loud one.
+    permissions: {}
     steps:
       # Board writes authenticate as the tracebloc-release-train App (backend#2036),
       # not a human's PAT. `owner:` yields an ORG-scoped installation token; a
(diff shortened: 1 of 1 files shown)


saadqbal:
/fr-pass

CI-only: scopes the add-to-kanban App token to least privilege. No product surface, so the staging journey neither covers it nor needs to — the repo's own CI is the verification. Advancing as part of the fleet-wide 2157/2166 sweep.

PR #516, opened 2026-08-20
chore(2212): drop the second version guard, keep the one that fails closed

LukasWodka:
## What

Deletes `.github/workflows/version-guard.yml`. This repo carried **two** version guards:

| Guard | Required? | Asserts |
|---|---|---|
| `version-bump-gate / version-check` | **yes**, on `develop` | one bump per **release cycle** — fails when a published path changes while the version file equals an already-released tag |
| `package change ⇒ __version__ bump` | **no**, nowhere | one bump per **PR** — fails when `tracebloc_ingestor/` changes and `__version__` wasn't bumped in that PR |

They are **not** equivalent, so this is not a like-for-like dedupe: mid-cycle the required gate deliberately says *"no bump is owed"* while the local guard still demands one. The real question is whether the local guard protects anything the required one doesn't. It doesn't.

## Proved, not argued

The required gate's logic is 409 lines of shell in `tracebloc/.github`, so it can be driven directly. I ran **its own script** against synthetic state:

| Scenario | Verdict | Gate's own message |
|---|---|---|
| version == latest released tag, published path changed, **no bump** | **FAIL** | `still says 0.8.10, v0.8.10 is ALREADY RELEASED, and this PR changes a published file` |
| mid-cycle (0.8.12 pending vs 0.8.10 released), no bump | pass | `is unchanged at 0.8.12 … v0.8.12 has NOT been released, so no bump is owed` |
| no published path touched | pass | `No publishable delta — 1 file(s), none matching [tracebloc_ingestor/*]` |
| bumped above latest | pass | `bumped 0.8.10 -> 0.8.11, and v0.8.11 is not released` |
| version regression below latest | **FAIL** | `says 0.8.9 …, v0.8.9 is ALREADY RELEASED` |

Each verdict names a **different branch** of the gate, which is what makes the harness non-vacuous. The first attempt at it *was* vacuous: a malformed compare stub failed all five cases identically at the merge-base check. That would have read as signal if I'd compared only exit codes rather than the messages.

Row 1 is exactly the **#408** shape the deleted guard was written for — the D16 `ds_<hex>` write path, merged without a bump, never reaching a released image. The required gate blocks it, because the first published change after any release finds `version == latest tag`. One bump per cycle is all it takes to cut a tag, so the per-PR strictness bought version churn and no additional safety.

Worth noting *why* a live test wouldn't have settled this: the repo is currently mid-cycle (`__version__ = "0.8.12"`, latest released `v0.8.10`), so a real PR would have hit row 2 and passed — inconclusive. Hence the harness.

## It was also the weaker of the two

By construction: it compared against the PR base, read the version with `grep`, and its own comment documents a latent fail-open — `grep -q` + `pipefail` → 141, read by `if !` as "no match" → `exit 0` "guard N/A", latent only because the path list is ~14 KB against a 64 KB pipe buffer (backend#1778). The required gate fails closed on an unreadable version file, an unlistable tag set, and a compare it cannot fully page.

## Checked before deleting

- `package change ⇒ __version__ bump` is a required context on **none** of `develop`, `staging`, `main`
- nothing outside the deleted file references it

## Reasoning preserved

Deleting the file shouldn't delete why it existed. The `#408` rationale and this proof now sit in `version-bump-gate-caller.yml`, next to the gate that carries it alone.

## Test plan

- Caller YAML parses
- Five-way harness above, driving the real gate script (not a reimplementation)
- CI on this PR exercises the surviving gate on a PR that touches no published path — expect `No publishable delta`

Refs tracebloc/backend#2212

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit b7191e00f28db3120b5e527284be3d6752e7fea2. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shujaatTracebloc (approved):
## Approving — the survivor genuinely subsumes the deleted guard, and I checked the part that could have gone wrong

Deleting a guard is the change that deserves the most suspicion, so I went after the claim that could have made this a coverage regression rather than a dedupe.

**The one thing that worried me: the two guards match paths differently.** The deleted guard used the regex `^tracebloc_ingestor/`, which matches at any depth. The survivor passes `publish-paths: "tracebloc_ingestor/*"` — and a single `*` in *pathname expansion* does not cross `/`. In this repo 61 of 72 package files sit one level deeper (`tracebloc_ingestor/ingestors/`, `/modalities/`, `/validators/`, …), so if that glob behaved like a filename glob the gate would miss ~85% of the package and this PR would be a real hole.

It doesn't. The gate matches with shell `case`, where pattern matching has no special treatment of `/`:

```
pat="tracebloc_ingestor/*"
tracebloc_ingestor/__init__.py              → MATCH
tracebloc_ingestor/ingestors/image.py       → MATCH
tracebloc_ingestor/modalities/a/b/x.py      → MATCH
tracebloc_ingestorX/y.py                    → no      (doesn't over-match either)
```

Same surface as the regex it replaces, in both directions. Worth stating explicitly in the PR body, incidentally — "matched with shell `case`" is in the reusable workflow's own docs, but a reader of *this* diff sees only `tracebloc_ingestor/*` and has to know that `case` and globbing differ to conclude it's safe.

**The required-context claim checks out exactly.** Read off branch protection rather than taken on trust:

```
develop  required: … "version-bump-gate / version-check"     ← present
staging  required: gate / gate, quality/*                     ← no version gate
main     required: pytest, e2e, gate / gate, quality/*         ← no version gate
```

`package change ⇒ __version__ bump` appears on none of the three, so it could never block a merge anywhere — which makes the case for removal stronger than the PR argues. It had social force (a red check), not enforcement.

**And the survivor does catch the #408 shape.** The branch is unambiguous in the reusable gate: `TAG_EXISTS=1` and `BASE_VER = HEAD_VER` and a publishable delta → `violation`, with the message the PR quotes. Since the first published change after any release necessarily finds `version == latest tag`, one bump per cycle is genuinely all it takes to cut a tag, and mid-cycle row 2 is a correct pass rather than a gap — a change landing under a pending `0.8.12` does reach a released image when that tag is cut.

**The fail-open you cite in the deleted guard is real**, and it's the right reason to prefer the survivor on top of redundancy: `grep -q` closing the pipe, `pipefail` turning SIGPIPE into 141, and `if !` reading that as "no match" → `exit 0`. Latent at ~14 KB against a 64 KB buffer, but latent is not absent, and the survivor fails closed on the same questions.

Good call keeping the `#408` rationale and the harness next to the gate that now carries it alone — deleting a guard shouldn't delete why it existed, and the caller comment is the right home for it.

One small observation, not a blocker: the deleted guard ran on `pull_request` with no branch filter, while the survivor is `branches: [develop]`. So a PR based directly on `staging`/`main` no longer gets it. That's fine for promotion PRs (no bump is owed) and it was advisory-only anyway — just noting the surface isn't quite identical, in case a direct hotfix to `main` is ever a path here.

CI green; the surviving gate reports `No publishable delta` on this PR as predicted.

PR #522, opened 2026-08-24
feat(telemetry): the ingest Job reports how every run ended (backend#2417)

LukasWodka:
## Summary

`data-ingestors` is one of the nine services registered in the platform telemetry contract
(RFC-BACKEND-1872 D2) — `service.name=data-ingestors`, `tracebloc.component=ingest-job` — and it
emitted **nothing**. It was never integrated and never had a ticket. Tracked as backend#2417.

The entry point now configures the shared contract emitter and reports one event per outcome on a
single object:

```
ingest.job.started  →  exactly one of  ingest.job.succeeded
                                       ingest.job.rejected   (config refused, nothing was read)
                                       ingest.job.failed
```

One object on purpose: it is what makes a failure **rate** computable rather than a bare count.

`main()` becomes the funnel and the run moves into `_run`. Every path out of `main()` reports
exactly one terminal event — including the exception nobody catches. That last case is not
theoretical: a missing DB credential or missing backend auth raises out of `Database()` /
`APIClient()` construction, and today that produces a raw traceback and no event at all. Paths that
classify themselves keep their own `error.type`; the funnel only fills silence.

The eight real endings each carry a closed, low-cardinality `error.type`:
`config_missing`, `config_not_found`, `config_unparseable`, `config_not_a_mapping`,
`config_invalid`, `ingestion_failed`, `records_failed`, plus `unhandled_exception` /
`unclassified` from the funnel.

## Redaction — the part that needed a decision

Ingestion touches customer cell values, and the contract forbids them outright. A rendered
traceback interpolates exactly one runtime value: **the exception message**. So the emitter is
handed `traceback.format_tb` — frames only, which are file, line, function and source line, i.e.
code — and never `format_exc` / `format_exception`.

That satisfies the contract's required half (`exception.type` + `exception.stacktrace`) exactly and
omits the field the contract already marks exempt (`exception.message`). Counts go out; rows do not.

## Where these records actually go — stated, not implied

This Job runs on the customer's cluster. The node-level collector takes control-plane containers
only, and widening it to ingestion is separate, blocked work. **So these records land in the
customer's own cluster logs and reach nothing centrally today.**

That is a real limitation and it is why this PR does not claim ingest failures are now centrally
visible. What it buys is conformance and local diagnosability now, and central visibility later as
a configuration change rather than a code change.

## Type of change

- [x] New feature (non-breaking)
- [x] Dependency change — `tracebloc-telemetry==0.1.0` added to `requirements.txt`

`tracebloc-telemetry` is the same pin the other Python services carry. Safe for a public,
pip-installable package: it is published on PyPI and declares **no** dependencies of its own; the
OpenTelemetry SDK is an optional `[otlp]` extra that is deliberately not taken, so without it every
contract rule is still enforced and records go to the standard library logger — which is the only
route out of this Job's process anyway.

## Test plan

**Ran (both CI matrix versions, Python 3.11 and 3.12):**

```
make check-all
  ruff check .                      All checks passed!
  pytest tests/ --cov-fail-under=95
  2139 passed, 1 xfailed            (was 2098)
  Required test coverage of 95% reached. Total coverage: 96.32%
    tracebloc_ingestor/telemetry.py    76 stmts   0 miss  100%
    tracebloc_ingestor/cli/run.py     119 stmts   2 miss   98%   (293-294, pre-existing)

black --check (26.3.1, the version code-quality pins)   5 files unchanged
gitleaks detect --config .gitleaks.toml                 no leaks found
```

**Not run, and why:** the `e2e (real MySQL)` suite needs a database this machine does not have. It
drives `run.main()`, whose signature and return values are unchanged, and it makes no assertions on
captured stdout/stderr (`grep -rn "capsys\|capfd\|caplog" e2e/` is empty). Its
`CLIENT_ENV=local` resolves to a registered non-exporting environment, so the emitter stays on the
console there by contract.

**Checks are derived, not restated.** No test here holds a second copy of the service table, the
domain list or the outcome verbs — they are read from the emitter's own registry and validated by
its own checkers. The `error.type` vocabulary is read off the entry point's **AST**, so a constant
declared and never used, a constant missing from the closed set, a call site passing a bare string,
and a rejection path that names no classification are all failures rather than review findings.

**Mutation-proved — ten mutations, each asserting its anchor applied before the run:**

| mutation | reddens |
|---|---|
| `format_tb` → `format_exception` (leak the message) | 4 tests |
| `_fail` stops classifying | 5 |
| the funnel's backstop removed | 3 |
| `ingest.job.failed` → an unregistered outcome verb | 11 |
| a constant declared but left out of the closed set | 2 |
| `exception.message` attached after all | 3 |
| the at-most-once flag never set (duplicate terminals) | 11 |
| environment read through `Config` (which defaults to prod) | 1 |
| a call site passes a bare string | 2 |
| a rejection path added with no classification | 2 |

**One of those mutations caught a vacuous assertion in these very tests**, which is the reason to
run them: `json.dumps` escapes non-ASCII, so the marker `patient-4711-Müller` sat in the record as
`Müller` and three end-to-end `SECRET not in dump` checks passed with the value right there.
Only the unit test reddened. They now assert on attribute values, and all four redden.

## Checklist

- [x] Ran the linter and the tests that cover the change
- [x] Read the full diff
- [x] No secrets, customer data, or customer-derived values in code, tests or messages
- [x] `CLAUDE.md` and `.cursor/BUGBOT.md` updated in the same PR (dependency list, module map, and
      a flag rule: a contract record is a wire, and `format_exc` in a telemetry path is a leak)
- [x] The stale `# pragma: no cover` on `main()` removed — it now carries the logic that
      guarantees a run says how it ended, so it is measured

---

> [!NOTE]
> **Medium Risk**
> Adds a new egress path for run outcomes from a customer-cluster Job. Redaction is the main risk (stack frames vs exception messages); ingest itself is unchanged if telemetry fails.
> 
> **Overview**
> Wires the ingest Job into the shared `tracebloc-telemetry` contract so every run reports `ingest.job.started` and exactly one of succeeded / rejected / failed, with a closed `error.type` vocabulary.
> 
> `main()` is now the funnel around `_run`: config refusals classify themselves, ingestion/record failures keep their type, and uncaught startup exceptions or unclassified non-zero exits are filled in rather than silent. Failure records send exception type and frames via `format_tb` only — never `exception.message` or failed rows, only counts. Telemetry is best-effort and cannot fail an ingest.
> 
> Adds `tracebloc-telemetry==0.1.0` (no OTLP extra) and bumps the package to `0.8.13`. Records currently stay in the customer cluster until collection is widened.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7d71fca3ddc4171385e0653bee1df39249be988. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/cli/run.py
+++ b/tracebloc_ingestor/cli/run.py
@@ -47,10 +47,12 @@ from typing import Any, Dict, Iterable, List, Optional
 import yaml
 from jsonschema import Draft7Validator, ValidationError
 
+from .. import telemetry
 from ..api.client import APIClient
 from ..config import Config
 from ..database import Database
 from ..ingestors import CSVIngestor, JSONIngestor
+from ..utils.correlation import resolve_correlation_id
 from ..utils.logging import setup_logging
 from .conventions import ResolvedConfig, resolve
 
@@ -68,12 +70,21 @@ _SCHEMA_PATH = Path(__file__).resolve().parent.parent / "schema" / "ingest.v1.js
 # ---------------------------------------------------------------------------
 
 
-def main(argv: List[str] | None = None) -> int:  # pragma: no cover - thin shell
+def main(argv: List[str] | None = None) -> int:
     """Entrypoint registered as the ``tracebloc-ingest`` console script.
 
     Returns the process exit code. A non-zero return is converted to
     ``sys.exit`` by the console-script wrapper; using a return value (rather
     than raising) keeps the function testable from inside pytest.
+
+    THIS IS THE TELEMETRY FUNNEL, and that is the only reason the run itself
+    lives in ``_run``. Every path out of this function — a clean exit, a
+    classified failure, an exception nobody caught — reports exactly one
+    terminal event, because the funnel reports one on every branch and
+    ``telemetry._terminal`` drops the duplicate when the path already
+    classified itself. A run that ends without saying how it ended is
+    indistinguishable from a run that never started, and that silence is the
+    thing worth removing.
     """
     # Group-writable by default (umask 002) so every directory and file the run
     # writes under DEST_PATH can be reclaimed by the `data delete` teardown
@@ -83,32 +94,68 @@ def main(argv: List[str] | None = None) -> int:  # pragma: no cover - thin shell
     # containing directories do, which this guarantees for the whole tree.
     os.umask(0o002)
 
+    telemetry.configure_job()
+    # Resolved here as well as in BaseIngestor, so the config-rejection paths
+    # below — which end the run before an ingestor exists — can still name their
+    # run. The only cost is that a MALFORMED id logs its warning twice, which is
+    # cheaper than a second copy of the id's validation rule.
+    telemetry.begin_run(resolve_correlation_id())
+
+    try:
+        code = _run(argv)
+    except BaseException as exc:
+        # Startup failures live here: a missing DB credential or missing backend
+        # auth raises out of Database() / APIClient() construction, and until now
+        # produced a raw traceback and no event whatsoever.
+        telemetry.job_failed(telemetry.ERROR_UNHANDLED_EXCEPTION, exc)
+        raise
+
+    if code == 0:
+        telemetry.job_succeeded()
+    else:
+        telemetry.job_failed(telemetry.ERROR_UNCLASSIFIED)
+    return code
+
+
+def _run(argv: List[str] | None = None) -> int:
+    """The run itself. See ``main`` for why it is a separate function."""
     config_path = os.environ.get("INGEST_CONFIG")
     if not config_path:
         return _fail(
             "INGEST_CONFIG env var not set. The official image expects the "
             "Helm subchart (client#86) to mount the ingest.yaml and set "
-            "INGEST_CONFIG to its path."
+            "INGEST_CONFIG to its path.",
+            telemetry.ERROR_CONFIG_MISSING,
         )
 
     raw_path = Path(config_path)
     if not raw_path.is_file():
-        return _fail(f"INGEST_CONFIG points to {config_path} which does not exist.")
+        return _fail(
+            f"INGEST_CONFIG points to {config_path} which does not exist.",
+            telemetry.ERROR_CONFIG_NOT_FOUND,
+        )
 
     try:
         raw_config = yaml.safe_load(raw_path.read_text(encoding="utf-8"))
     except yaml.YAMLError as e:
-        return _fail(f"ingest.yaml is not valid YAML:\n  {e}")
+        return _fail(
+            f"ingest.yaml is not valid YAML:\n  {e}",
+            telemetry.ERROR_CONFIG_UNPARSEABLE,
+        )
 
     if not isinstance(raw_config, dict):
         return _fail(
             "ingest.yaml must be a mapping at the top level "
-            "(apiVersion / kind / category / ...)."
+            "(apiVersion / kind / category / ...).",
+            telemetry.ERROR_CONFIG_NOT_A_MAPPING,
         )
 
     errors = list(_validate(raw_config))
     if errors:
-        return _fail("ingest.yaml validation failed:\n" + _format_errors(errors))
+        return _fail(
+            "ingest.yaml validation failed:\n" + _format_errors(errors),
+            telemetry.ERROR_CONFIG_INVALID,
+        )
 
     resolved = resolve(raw_config)
 
@@ -166,12 +213,18 @@ def main(argv: List[str] | None = None) -> int:  # pragma: no cover - thin shell
             # the exception into a clean, single-line non-zero exit rather than a
             # raw traceback, so the CLI's live log stream shows an actionable
             # reason and marks the Job failed.
+            telemetry.job_failed(telemetry.ERROR_INGESTION_FAILED, exc)
             print(f"\nIngestion failed: {exc}", file=sys.stderr)
             return 1
         if failed:
             logger.warning(
                 "%d record(s) failed during ingestion; see logs for details.",
                 len(failed),
+            )
+            # The COUNT, never the records: counts are dataset structure and may
+            # be reported, the rows themselves may not.
+            telemetry.job_failed(
+                telemetry.ERROR_RECORDS_FAILED, failed_records=len(failed)
             )
             return 1
 
@@ -369,11 +422,18 @@ def _build_ingestor(
 # ---------------------------------------------------------------------------
 
 
-def _fail(message: str) -> int:
+def _fail(message: str, error_type: str) -> int:
     """Print to stderr and return non-zero. Logger may not be configured yet
     when validation fails (it depends on Config() which depends on env
-    being set), so a plain stderr write is the most reliable channel."""
+    being set), so a plain stderr write is the most reliable channel.
+
+    The single funnel for "the configuration was refused and no data was read",
+    so a rejection path cannot be added without also classifying itself.
+    ``error_type`` is a constant from ``telemetry.ERROR_TYPES``; a literal here
+    is a finding, and ``tests/test_telemetry.py`` reads this module to say so.
+    """
     print(message, file=sys.stderr)
+    telemetry.job_rejected(error_type)
     return 2
 
 
--- a/tracebloc_ingestor/cli/run.py
+++ b/tracebloc_ingestor/cli/run.py
@@ -40,6 +40,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import signal
 import sys
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional
@@ -100,6 +101,26 @@ def main(argv: List[str] | None = None) -> int:
     # run. The only cost is that a MALFORMED id logs its warning twice, which is
     # cheaper than a second copy of the id's validation rule.
     telemetry.begin_run(resolve_correlation_id())
+
+    def _on_sigterm(signum, _frame):  # pragma: no cover - driven by a real signal
+        # A run the platform ended is still a run that ended, and SIGTERM's
+        # default disposition unwinds NOTHING -- no exception, no `finally`, no
+        # `atexit` -- so the terminal event is emitted here or not at all. The
+        # funnel below catches BaseException, which covers every way Python
+        # leaves the frame and none of the ways a signal does. Measured on this
+        # branch by @shujaatTracebloc: SIGTERM at 3s gave `started` and silence.
+        #
+        # In Kubernetes this is the common ending, not an edge case: eviction,
+        # node drain and `activeDeadlineSeconds` all send SIGTERM. SIGKILL
+        # cannot be caught, so an OOMKilled run still reports `started` alone —
+        # `_terminal`'s docstring says so rather than implying otherwise.
+        #
+        # `_terminal` makes this safe against double-reporting if the signal
+        # lands while a classified path is already mid-flight.
+        telemetry.job_cancelled()
+        raise SystemExit(128 + signum)
+
+    signal.signal(signal.SIGTERM, _on_sigterm)
 
     try:
         code = _run(argv)
--- a/.cursor/BUGBOT.md
+++ b/.cursor/BUGBOT.md
@@ -37,6 +37,8 @@ PyPI first. `RELEASING.md` documents how they stay aligned.
   suppressed" note (`ingestors/csv_ingestor.py:508`,
   `validators/per_group_time_ordered_validator.py:219`, `metadata_backfill_runner.py:195`).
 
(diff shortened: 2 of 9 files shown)


tracebloc_ingestor/cli/run.py
```diff
@@ -83,32 +94,68 @@ def main(argv: List[str] | None = None) -> int:  # pragma: no cover - thin shell
...
+    telemetry.configure_job()
+    # Resolved here as well as in BaseIngestor, so the config-rejection paths
+    # below — which end the run before an ingestor exists — can still name their
+    # run. The only cost is that a MALFORMED id logs its warning twice, which is
+    # cheaper than a second copy of the id's validation rule.
+    telemetry.begin_run(resolve_correlation_id())
+
+    try:
+        code = _run(argv)
+    except BaseException as exc:
```
shujaatTracebloc, line 127: **SIGTERM leaves exactly the silence this PR exists to remove — measured.**

`except BaseException` covers every way Python leaves this frame, so the funnel's claim holds for the *language*. It does not hold for the signal, and in Kubernetes the signal is the common case: pod eviction, node drain and `activeDeadlineSeconds` all send SIGTERM, whose default disposition terminates the interpreter without unwinding — no exception, no `finally`, no `atexit`.

Driven against this branch with `_run` stubbed to sleep, then SIGTERM'd at 3s:

```
telemetry: not exporting — environment '<unset>' ...
ingest.job.started service.name=data-ingestors ... severity=INFO
RUN-STARTED
                                        <- process gone, exit 143
terminal events emitted: 0
```

So an evicted ingest emits `started` and nothing else — "a run that ends without saying how it ended", which the docstring three lines above names as the thing worth removing. It is also the failure mode where the record would be *most* useful: an OOM-killed or deadline-exceeded ingest is precisely what nobody can currently count.

Non-blocking — emitting `started`-only is still strictly better than today's nothing, and I can see the argument that signal handling is its own change. Two things make me raise it rather than let it pass:

1. The pairing rule is described as structural ("never zero"), and this is a reachable path where it is zero. If it stays, the docstring is the thing to adjust — say the guarantee covers Python-level exits and that signal termination is a known gap, so the next reader doesn't trust it further than it goes.
2. The emitter's registry already has the vocabulary for it. `OUTCOMES` includes `cancelled` and `timed_out`, so `ingest.job.cancelled` validates today — no registry change needed.

The minimal shape, if you want it in this PR:

```python
    telemetry.configure_job()
    telemetry.begin_run(resolve_correlation_id())

    def _on_sigterm(signum, _frame):
        # A Job that is evicted, drained or deadline-exceeded still ended, and
        # the default SIGTERM disposition unwinds nothing -- so the terminal
        # event has to be emitted here or not at all.
        telemetry.job_cancelled()
        raise SystemExit(128 + signum)

    signal.signal(signal.SIGTERM, _on_sigterm)
```

`_terminal` already makes this safe against double-reporting if the handler fires while a classified path is mid-flight.

shujaatTracebloc (commented):
## One non-blocking note; everything else I could check, I checked, and it holds

I'm commenting rather than approving for a single reason: the inline note is a reachable path where the pairing rule is zero, and I'd rather you decide whether it belongs in this PR than have my approval imply I hadn't noticed. It is **not** a blocker — `started`-only on eviction is strictly better than today's nothing.

### Verified, not read

- **The redaction claim is exactly true, and it is pinned.** With a `ValueError("cannot cast 'PATIENT-SSN-123-45-6789' to int")` raised through two frames: `safe_stacktrace` yields the frame list with the value absent, `traceback.format_exc()` on the same exception contains it, and the emitted record carries `error.type`, `exception.type=ValueError` and `exception.stacktrace` with no trace of the value. Swapping `format_tb` for `format_exception` reddens `test_telemetry_failure_record_never_carries_the_exception_message` and `test_a_failed_ingestion_reports_frames_but_never_the_exception_message`. Sending the contract's required half and omitting the exempt `exception.message` is the right reading, and `format_tb` is the right instrument for it.
- **Five mutations, five caught**, on my own registry rather than the description's:

  ```
  drop the _terminal dedup                  -> 11 failed
  safe_stacktrace uses format_exception     ->  4 failed  (incl. the redaction test)
  drop job_failed(ERROR_UNCLASSIFIED)       ->  3 failed
  _fail stops calling job_rejected          ->  5 failed
  drop the escaping-exception report        ->  3 failed
  ```

  `test_main_reports_exactly_one_terminal_event` parametrised over *returns_zero / returns_nonzero / raises* is the right shape — that is the whole of `_run`'s signature, so the guarantee is checked over its actual domain rather than at sampled points.
- **Every constant validates against the live registry**, installed at the pinned `tracebloc-telemetry==0.1.0`: `data-ingestors` is a registered service, `ingest-job` is in its component set, and all four `ingest.job.*` names resolve to a registered domain and outcome with `failed`/`rejected` correctly in `FAILURE_OUTCOMES`. Every one of the nine `ERROR_TYPES` emits, `tracebloc.ingest.failed_records` is accepted, and the module's own warning logger stayed empty throughout — so nothing this code sends is being silently refused into a `could not emit` line.
- **The environment reasoning is right and the degradation is loud.** `CLIENT_ENV` unset → the emitter announces "not exporting — environment '\<unset\>' is not one of dev/prod/stg — refusing to guess one" and *still* emits contract-shaped records to the console; `CLIENT_ENV=prod` → announces the missing OTLP endpoint and stamps `deployment.environment=prod`. Reading the raw var rather than `Config.EDGE_ENV` is the correct call for exactly the reason given — `EDGE_ENV`'s `"prod"` default is right for picking an endpoint and wrong for stamping a record — and the installers canonicalise `CLIENT_ENV` to `dev|stg|prod` (backend#1745), which is precisely `EXPORTING_ENVIRONMENTS`.
- **The funnel classifies every path.** All six config rejections go through `_fail` with a constant, both run failures classify themselves, and `_terminal` drops the funnel's backstop duplicate. `template_runner`'s `sys.exit(1)` is not reachable from `run.py`, and `_run` takes no argparse, so there is no stray `SystemExit` to be misfiled as `unhandled_exception`.
- **Suite green:** 2139 passed, 1 xfailed. Version bumped 0.8.12 → 0.8.13 in lockstep, and the added dep is justified in the requirements comment rather than just pinned.

Setting the terminal flag *before* the emit is the subtle call I'd have wanted argued, and the docstring argues it correctly: a run whose terminal record could not be delivered still ended, and letting the backstop fire behind it would report one ending twice under two classifications.

Two things I looked at and am deliberately not raising: `job_failed`'s attribute assembly sits outside the `_emit` guard, so "every call here is guarded" is slightly stronger than the code (but I can't construct a realistic input where `format_tb` or `int(len(...))` raises); and `except BaseException` catching `KeyboardInterrupt` as `unhandled_exception`, which is defensible and not reachable in a Job.

tracebloc_ingestor/cli/run.py
shujaatTracebloc: Re-review picked this up, but the head is still `e7d71fc` — the same commit I reviewed — with no reply here, so I've nothing new to weigh and I'm leaving this open rather than resolving it on my own initiative.

To be clear about what I'm asking for, because it is deliberately small: **this was never a blocker and I am not asking you to implement signal handling.** Either of these closes it and I'll approve on the next pass:

1. A one-line "won't fix in this PR" here — I'll take it, and I'd suggest a follow-up ticket so the gap is recorded somewhere other than a resolved review thread. A considered decline is yours to make.
2. Or narrow the claim in `main`'s docstring: it currently says "Every path out of this function ... reports exactly one terminal event", and the measured SIGTERM case is a path where it reports zero. Something like "every path Python can leave this frame by" makes the sentence true, and leaves the next reader knowing where the guarantee ends.

If you'd rather I just resolved it and approved, say so and I will — I only want the decision to be yours and visible, not inferred from a re-request.

LukasWodka: Fixed in `6e7fa74`. You measured it rather than reasoned about it, and that is why it is fixed rather than documented away — thank you.

You offered two options; I took the second, because option 1 would have left the reachable zero in place and only lowered the docstring's claim to match it. The vocabulary was already there, as you said: `cancelled` is in the emitter's closed `OUTCOMES`, so `ingest.job.cancelled` validates with no registry change.

Shipped close to your shape:

- SIGTERM handler emitting `ingest.job.cancelled`, added to `TERMINAL_EVENTS`.
- **Distinct from `job_failed` deliberately** — nothing about the ingest went wrong, and counting evictions as failures would make the failure rate a function of cluster pressure rather than of ingestion.
- Two tests, both driving **the real handler `main` installs** rather than a copy of it, so deleting the `signal.signal(...)` line reddens them. One asserts exactly one terminal event on the signal path; the other asserts a signal landing *after* a classified path is **dropped**, not double-reported — pinning the `_terminal` property you pointed at, for the path that runs outside every control-flow guarantee the funnel relies on.

**One refinement to your suggestion, and it is the reason I did not just paste it.** Your note mentions an OOM-killed ingest. **SIGKILL cannot be caught**, so an OOMKilled run still emits `started` alone — the handler covers eviction, drain and `activeDeadlineSeconds`, which send SIGTERM first, and does not cover OOMKilled, which does not. Taking the shape without that distinction would have moved the over-claim rather than removed it.

So `_terminal`'s docstring now scopes its guarantee explicitly — Python-level exits **and** SIGTERM, **not** SIGKILL — instead of saying "structurally, never zero" and being wrong in a newer, harder-to-spot place. Your point 1 turned out to be right even after taking option 2: the docstring did need adjusting, just less far than it would have.

shujaatTracebloc: Verified on `6e7fa74`, and this is a better outcome than either option I offered — you fixed it rather than declining it or narrowing the prose, and then narrowed the prose as well.

I re-ran the exact experiment that produced the finding: `_run` stubbed to sleep, SIGTERM at 3s.

```
before (e7d71fc):  ingest.job.started ... RUN-STARTED       exit 143, terminal events: 0
after  (6e7fa74):  ingest.job.started ...
                   ingest.job.cancelled ... severity=WARN   exit 143, terminal events: 1
```

`ingest.job.cancelled` validates against the registry without a change (`cancelled` is in `OUTCOMES`, and correctly *not* in `FAILURE_OUTCOMES`), and it's in `TERMINAL_EVENTS`, so the pairing rule covers it rather than treating it as a fifth spelling.

The tests drive the real handler, which is the part that could have been faked — deleting `signal.signal(signal.SIGTERM, _on_sigterm)` from `main` reddens **both** new tests, and routing `job_cancelled` around `_terminal` reddens the double-report one specifically. 2141 pass.

Two judgement calls worth naming because they're better than what I asked for: making `cancelled` distinct from `failed` rather than reusing `job_failed` — "counting these as failures would make the failure rate a function of cluster pressure" is the right reason and one I hadn't articulated — and stating in `_terminal`'s docstring that **SIGKILL cannot be caught**, so an OOMKilled run still reports `started` alone. Naming the residue beats implying it's gone.

Satisfied. Approving.

shujaatTracebloc (approved):
## Approving — the SIGTERM gap is fixed in code, verified end to end

You took the harder of the two options I offered and it was the right one.

**The finding is closed by measurement, not by concession.** Same experiment that produced it — `_run` stubbed to sleep, SIGTERM at 3s:

```
before (e7d71fc):  started ... RUN-STARTED                  exit 143, terminal events: 0
after  (6e7fa74):  started ... cancelled (severity=WARN)    exit 143, terminal events: 1
```

`ingest.job.cancelled` validates against the live registry with no registry change — `cancelled` is in `OUTCOMES` and correctly absent from `FAILURE_OUTCOMES` — and it is in `TERMINAL_EVENTS`, so the pairing rule extends to it rather than a fourth ending sitting outside the invariant.

**The new tests bite where it counts.** They drive the handler `main` actually installs rather than a copy, which is the thing that could have been vacuous here:

```
delete signal.signal(SIGTERM, _on_sigterm) from main  -> 2 failed  (both new tests)
job_cancelled bypasses _terminal                      -> 1 failed  (the double-report test)
```

2141 passed, 1 xfailed. All checks green, Bugbot included.

Two calls better than what I asked for. Keeping `cancelled` distinct from `failed` — "counting these as failures would make the failure rate a function of cluster pressure" — is the right reason and one I hadn't articulated; folding it into `job_failed` would have quietly made the failure rate track node churn. And narrowing `_terminal`'s docstring to say outright that **SIGKILL cannot be caught**, so an OOMKilled run still emits `started` alone, is the honest version: the guarantee now says where it ends instead of implying it has no end.

Everything from my earlier pass still stands — the redaction claim holds against a planted secret, `format_exc` would have leaked it, five mutations all caught, every constant validates against the registry, and the `CLIENT_ENV` reasoning matches the installers' canonical `dev|stg|prod`. Nothing outstanding from me.

PR #524, opened 2026-08-24
fix(telemetry): a signal after the load is durable is a success (backend#2435)

LukasWodka:
Closes tracebloc/backend#2435

Fixes the Bugbot Medium shipped at the `develop -> staging` hop on #523
([thread](https://github.com/tracebloc/data-ingestors/pull/523#discussion_r3842395723)),
tracked as backend#2435. **The finding was real** and reproduced with a real signal
before anything was changed.

## The state machine as I found it

`main` is the telemetry funnel; `_run` is the run. Of the five events, four are
terminal, and `_terminal` is **first-writer-wins**.

| Outcome | Emitted where | Classified at the moment it becomes true? |
|---|---|---|
| `rejected` (5 config paths) | inside `_fail`, before returning | yes |
| `failed` / `ingestion_failed` | inside `_run`'s `except` | yes |
| `failed` / `records_failed` | inside `_run`, before returning | yes |
| `failed` / `unhandled_exception` | the funnel, already unwinding | yes |
| **`succeeded`** | **the funnel, after `_run` returned** | **no** |
| `cancelled` | the SIGTERM handler, whenever it fires | n/a |

Success was the one late-classifying outcome, so it was the one a signal could
overwrite. The `cancelled` path is not a rare ending either — in Kubernetes,
eviction, node drain and `activeDeadlineSeconds` all send SIGTERM.

**The window is wider than the finding's wording suggests.** `reclaim_source` is
not in `__exit__` (that method is `pass`) — it runs *inside* `ingest`, at
`base.py`'s `if dataset_registered and not failed_records:` guard, and it is an
`rmtree` over a whole staged dataset. Everything after it — the table-lock
release in `ingest`'s `finally`, the context manager's exit, `logger.info`,
`_run`'s return, the funnel's `if code == 0` — is also inside the window. I
checked what can still *raise* in there: `reclaim_source` swallows every
exception by contract, and `TableLock.release` swallows `FileNotFoundError` and
logs anything else. So nothing after that guard can change the outcome — which
is what makes claiming success there sound.

## Reproduced before fixing

Real `kill(SIGTERM)` through the entrypoint's own installed handler, on `develop`
at `c795af5` (the commit Bugbot reviewed, and still the tip of `develop`):

```
[repro] load durable: rows committed + dataset registered
[repro] exited 143
EVENTS:   ['ingest.job.started', 'ingest.job.cancelled']
TERMINAL: ['ingest.job.cancelled']
```

A committed, registered, live dataset reported as cancelled.

## The mapping, before and after

| Signal arrives… | before | after |
|---|---|---|
| during validation / the load | `cancelled` | `cancelled` (unchanged) |
| during `reclaim_source`'s rmtree, inside `ingest` | `cancelled` | `succeeded` |
| after `ingest` returns clean, before the funnel | `cancelled` | `succeeded` |
| after the funnel already reported | dropped | dropped (unchanged) |
| never (clean run) | `succeeded` | `succeeded` (unchanged) |
| after durability, but something then failed and said so | `failed` | `failed` (unchanged) |

## Why a flag and not an early `job_succeeded()`

`mark_durable()` records the fact and leaves the terminal slot **open**. Emitting
`succeeded` at the durability point would have closed it, and a genuine failure
during the teardown would then be dropped as a duplicate — a broken run
reporting success. That is the expensive direction of this bug, so it is not
traded for the cheap one. `test_durability_does_not_swallow_a_later_genuine_failure`
pins that, and mutation M8 is the proof it would catch the collapse.

Two call sites, both idempotent: `BaseIngestor` claims durability at the guard it
already uses to decide the load is durable and clean (covering the rmtree), and
the entrypoint claims it again as a backstop for any success path that returns no
failed records without reaching that proof point.

## Tests

Ten new cases. The two end-to-end ones drive the **real** `main()` / `_run`
through `test_cli_run.py`'s existing mocked-runtime harness and deliver a **real**
`kill`, so the assertions are coupled to production's `mark_durable` call and not
to a copy of it in the test.

The vocabulary is derived, not restated: `_ESTABLISH` is keyed by outcome and
`test_every_establishable_outcome_is_covered` compares it against
`telemetry.TERMINAL_EVENTS`, naming `cancelled` as the one deliberate absence — so
a fifth terminal event fails that test instead of silently going unexercised.

## Mutation table

Every anchor asserted to occur exactly once before replacing (an inert mutation
and real coverage look identical in a log), then restored. Baseline 69 passed.

| # | Mutation | Test that reddened |
|---|---|---|
| M1 | `run.py`: entrypoint's `mark_durable()` deleted | `test_a_signal_during_teardown_reports_the_success_it_already_earned` |
| M2 | `job_cancelled` ignores durability (**the shipped bug**) | that test + `…outcome_already_established[succeeded]` |
| M3 | `job_cancelled` *always* reports success | `test_a_signal_before_the_work_is_done_still_reports_a_cancellation`, `test_sigterm_reports_the_run_as_cancelled` (pre-existing), both `…next_run` cases |
| M4 | `begin_run` stops clearing `_durable` | `test_durability_does_not_survive_into_the_next_run[begin_run]` |
| M5 | `reset` stops clearing `_durable` | `…[reset]` |
| M6 | `base.py`: ingestor stops claiming durability | `test_the_ingestor_claims_durability_before_it_reclaims_the_source` |
| M7 | `base.py`: durability claimed *after* the reclaim | same (order is asserted, not just presence) |
| M8 | `mark_durable` collapsed into an early `job_succeeded()` | `test_durability_does_not_swallow_a_later_genuine_failure` + `…[succeeded]` |

**M5 was vacuous on first run and the test was strengthened, not the log.**
`reset()` was followed by `begin_run()`, which also clears the flag, so deleting
`reset`'s clear reddened nothing — the case was proving what `begin_run` does
twice and what `reset` does not at all. It now asserts `durable() is False`
immediately after `reset()`, and M5 discriminates.

## Verification

- `python -m py_compile` on all five changed files — OK
- `black --check` at the pinned **26.3.1** (`code-quality-caller.yml`) — 5 files unchanged
- CI's own gate, `make coverage` (= `pytest tests/ --cov-fail-under=95`):
  **2151 passed, 1 xfailed, 96.34%**. Baseline on `origin/develop` was 2141 passed,
  so the +10 are exactly the new cases. `telemetry.py` is at 100%.

Note: `tracebloc-telemetry==0.1.0` is not published to the index this machine can
reach, so it was supplied from the in-repo source tree on `PYTHONPATH`. CI installs
it from `requirements.txt` normally.

## Deliberately out of scope

**The exit code is unchanged** — a signalled run still exits 143, so its pod is
still a failed pod even when telemetry now reports `succeeded`. Mapping it to 0
would change Kubernetes Job retry semantics and deserves its own change and its
own functional review; this PR fixes what the run *reports*, which is what was
being undercounted.

`SIGKILL` remains uncovered and still emits `started` alone — it cannot be
caught, exactly as `_terminal`'s docstring already says.

## Test plan

- [x] Misreport reproduced on `develop` with a real signal before any change
- [x] `pytest tests/` green with the coverage gate CI applies
- [x] Every added test mutation-proven, with the anchor count asserted
- [x] `black --check` at 26.3.1
- [x] `CLAUDE.md`'s telemetry entry updated in the same PR (`mark_durable` is now
      called from `ingestors/base.py`, while the entry point stays the only emitter)

---

> [!NOTE]
> **Medium Risk**
> Changes how ingest jobs classify SIGTERM vs success in contract telemetry. Wrong durability timing would miscount Kubernetes evictions or hide a post-commit failure; ingest data path and exit codes are unchanged.
> 
> **Overview**
> Fixes backend#2435: a SIGTERM during post-commit teardown (source reclaim, lock release) no longer steals the first-writer-wins terminal slot as `ingest.job.cancelled` for a dataset that is already committed and registered.
> 
> Adds `mark_durable()`, which records that fact and **does not emit**. `job_cancelled()` then reports `succeeded` if durable, otherwise `cancelled`. The slot stays open so a real failure after commit still classifies itself. Called from `BaseIngestor` immediately before `reclaim_source`, and again from `_run` as an idempotent backstop. `begin_run` / `reset` clear the flag so it cannot leak into a later run.
> 
> Exit code is still 143 on signal. Tests cover real `kill` during teardown vs mid-ingest, no overwrite of established outcomes, later genuine failure, and call order before reclaim.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 569a848b3ac183e9fce662e7f0bb0d2ce9eb16db. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shujaatTracebloc (approved):
## Approving — no findings. And this is the second-order cost of the handler I asked for on #522.

Worth saying plainly: I pushed for the SIGTERM handler on #522 and approved it. Before that, a signalled run emitted `started` and nothing else; after it, the same run emits `cancelled` — including when the rows were already committed and the dataset registered. So I traded silence for a wrong classification, and this PR is the correction. Bugbot caught it at the promotion hop; I didn't, on a PR where I'd read the `_terminal` docstring closely enough to quote it back. The first-writer-wins property was right there and I didn't ask which outcome classifies late.

### Verified, not read

**The behaviour matrix is exactly right**, driven directly:

```
not durable + SIGTERM        -> ingest.job.cancelled
durable     + SIGTERM        -> ingest.job.succeeded     <- the fix
durable     + genuine failure-> ingest.job.failed        <- the expensive direction, protected
durable     + rejection      -> ingest.job.rejected
```

That third row is the one that matters. Recording a fact instead of emitting `succeeded` early is the right call and for the reason you give: emitting would close the terminal slot, and a teardown failure would then be dropped as a duplicate — a broken run reporting success. Leaving the slot open costs nothing and keeps the expensive direction closed.

**Durability doesn't leak between runs.** `mark_durable()` → `True`; `begin_run()` → `False`; `reset()` → `False`. Both clearers matter and both are guarded.

**Five mutations, five caught** — my own, not the table's:

```
M2  job_cancelled ignores durability (the shipped bug)  -> 2 failed, incl. the end-to-end teardown case
M8  mark_durable also emits succeeded early             -> test_durability_does_not_swallow_a_later_genuine_failure
M7  durability claimed AFTER the reclaim                -> test_the_ingestor_claims_durability_before_it_reclaims_the_source
M4  begin_run stops clearing _durable                   -> …does_not_survive_into_the_next_run[begin_run]
M5  reset stops clearing _durable                       -> …[reset]
```

M7 is the one I'd have worried about — order, not presence — and it genuinely discriminates, so the rmtree window is actually covered rather than incidentally. And **M5 discriminates now**, which is the interesting one: you found it vacuous because `reset()` was followed by `begin_run()` (which also clears), fixed the test rather than the log, and it bites. That is the same class you caught on #213's `model_type` entry — an inert mutation and real coverage looking identical. Twice now, self-caught.

**The vocabulary is derived.** Adding a fifth terminal event (`ingest.job.timed_out`) reddens `test_every_establishable_outcome_is_covered`, so a new ending can't slip in unexercised, with `cancelled` named as the one deliberate absence.

Suite 2151 passed, 1 xfailed — exactly your baseline of 2141 plus the ten new cases.

The analysis in the description earns its length, particularly correcting the finding's own wording: `reclaim_source` is not in `__exit__`, it's inside `ingest` at the `dataset_registered and not failed_records` guard, and an rmtree over a staged dataset is the long part of the window rather than a theoretical one. Deriving the claim from the same guard the reclaim is derived from is what stops the two disagreeing about what "succeeded" means.

PR #527, opened 2026-08-24
fix(token-classification): count distinct BIO tags, not tag sequences (backend#1747)

aptracebloc:
Closes tracebloc/backend#1747

## Problem

`token_classification` stores the whole per-row **BIO tag SEQUENCE** (one tag per word, e.g. `"O B-PER I-PER O"`) in the `label` column. The ingest-summary count path ran `GROUP BY label` (`Database.get_label_counts`) over that column, so `output_classes` became the set of **distinct sequence strings** (e.g. 77) instead of the handful of **distinct tags** (e.g. 9) the model head is built on. No head links to that class set — the whole task type is unrunnable end to end.

## Fix

- **`Database.get_tag_counts`** — new count path that explodes each row's whitespace-joined tag sequence into individual tags and tallies token occurrences per tag. The `GROUP BY label` still runs in SQL, so the work is `O(distinct sequences)`, not `O(rows)`.
- **`ModalitySpec.label_is_tag_sequence`** — new trait, set on `token_classification` in the registry. `ingestors/base.py` selects `get_tag_counts` over `get_label_counts` **trait-style, never on a category string** — mirroring the existing `grouping` trait — so a future tag-sequence category is a one-line registry entry.
- The `counts_helper` error-label is now set once per branch (removes a duplicated ternary and stays truthful for the new branch).

This ports the tag-explosion logic from the deleted backend `global_meta/dataset_metadata_json_generator._explode_tags` into the ingestor's count path (that file is **not** restored). The CLI layout contract is untouched — the trait drives the count only, not on-disk staging.

## Existing datasets

Existing BIO datasets pick up correct `output_classes` on **re-ingest** through this path. The metadata-backfill runner (`metadata_backfill*.py`) deliberately does not carry row-level `labels` (its payload is schema + feature_stats only), so it is not extended here — correcting a stored `output_classes` without a re-ingest is a backend-endpoint change, out of this repo's scope.

## Testing

- New `tests/test_token_classification_tag_counts.py` (8 tests): the registry trait (set + unique + absent from the layout contract); base.py routing (token_classification ships distinct-tag counts, a non-tag text category keeps row counts); `get_tag_counts` explode-and-weight arithmetic (the 77→9 collapse), NULL-label handling, and SQL/escaping shape.
- Full suite green: **2159 passed, 1 xfailed**. `black --check` and `ruff check` clean.

## Known trade-off — backend record counts (accepted, tracked)

Thanks to @LukasWodka and @shujaatTracebloc for catching this. The `labels` dict does two jobs: its **keys** are the class vocabulary (fixed here), but the backend also infers a **record count** by summing its **values** (`sum(labels.values())`). Before this change that sum equalled the row count for `token_classification`; after it, it's the tagged-token count — so the backend's record-count fields (`total`, `unique_data_items_count`) inflate ~mean-tokens-per-row for this modality, on the single-edge path **and**, compounding, through `combine_datasets`.

Per-field: `count.label` becomes **right** (the point of this fix), `label_density` **better**, while `total` / `unique_data_items_count` become **wrong**. The class fix is the far more valuable half — an unrunnable task type is the worse failure — so this **ships with the inflated counts as a known, accepted consequence**, not a silent one. The real fix is cross-repo (carry an explicit row count in the summary payload and consume it instead of the sum for tag-sequence categories) and is **recorded on backend#1747** ([decision + full trace, both corrupted paths named](https://github.com/tracebloc/backend/issues/1747#issuecomment-5406738806)) as the follow-up.

---

> [!NOTE]
> **Medium Risk**
> Changes the ingest-summary class set for token classification, which drives backend `output_classes` and model-head linking. Scope is one modality and is covered by unit tests, but a wrong explode would still ship a bad class vocabulary.
> 
> **Overview**
> Fixes token classification ingest so `output_classes` are the distinct BIO **tags**, not distinct per-row tag **sequence strings** (backend#1747). That mismatch left the model head unlinked and the task unrunnable.
> 
> Adds `Database.get_tag_counts`, which still `GROUP BY`s sequences in SQL then explodes and weights tags in Python (`O(distinct sequences)`). A new `ModalitySpec.label_is_tag_sequence` trait (set only on `token_classification`) selects this helper in the ingest-summary path—trait-style, not a category string. The CLI layout contract is unchanged.
> 
> NULL labels contribute no tags. Existing datasets get correct classes on re-ingest; metadata backfill is out of scope.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea44b5b36fc66d1c0a4c8cb0cb67857897ec495f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tracebloc_ingestor/ingestors/base.py
+++ b/tracebloc_ingestor/ingestors/base.py
@@ -527,6 +527,19 @@ class BaseIngestor(ABC):
         return spec.grouping if spec is not None else None
 
     @property
+    def _label_is_tag_sequence(self) -> bool:
+        """Whether the category's ``label`` column holds a whitespace-joined
+        per-token tag SEQUENCE (token_classification BIO/IOB2), so its
+        output_classes are the DISTINCT TAGS — the sequence exploded — rather
+        than the distinct sequence strings a plain ``GROUP BY label`` would
+        count (backend#1747). Read trait-style from the registry, never via a
+        category string, so a future tag-sequence category is a registry entry,
+        not a base.py edit. Selects ``get_tag_counts`` over ``get_label_counts``
+        in the ingest-summary count path."""
+        spec = _MODALITY_REGISTRY.get(self.category)
+        return bool(spec is not None and spec.label_is_tag_sequence)
+
+    @property
     def _table_lock(self) -> TableLock:
         """The run's table lock (P5b). ``TableLock`` owns the file-lock
         lifecycle — compute path, atomic acquire with stale-reclaim, release —
@@ -1200,6 +1213,7 @@ class BaseIngestor(ABC):
                 # future grouped category counting rows falls through to the
                 # standard row counts like every non-grouped category.
                 if grouping is not None and grouping.count_unit == "sequences":
+                    counts_helper = "get_label_sequence_counts"
                     label_counts = self.database.get_label_sequence_counts(
                         self.physical_table_name,
                         self.ingestor_id,
@@ -1212,7 +1226,19 @@ class BaseIngestor(ABC):
                     self.file_options["number_of_sequences"] = sum(
                         label_counts.values()
                     )
+                elif self._label_is_tag_sequence:
+                    # token_classification: the ``label`` column holds the whole
+                    # per-token BIO tag SEQUENCE, so output_classes are the
+                    # DISTINCT TAGS. Explode the sequence (get_tag_counts) rather
+                    # than GROUP BY the raw sequence string, which would count
+                    # distinct sequences as classes — no model head links then,
+                    # and the task is unrunnable e2e (backend#1747).
+                    counts_helper = "get_tag_counts"
+                    label_counts = self.database.get_tag_counts(
+                        self.physical_table_name, self.ingestor_id
+                    )
                 else:
+                    counts_helper = "get_label_counts"
                     # CEILING (review on #487, tracked in #488): this GROUP BYs
                     # the RAW label, so a regression-class dataset with a
                     # continuous target yields up to one entry per distinct
@@ -1247,11 +1273,6 @@ class BaseIngestor(ABC):
                         "skipping ingest summary."
                     )
                 elif not label_counts:
-                    counts_helper = (
-                        "get_label_sequence_counts"
-                        if grouping is not None and grouping.count_unit == "sequences"
-                        else "get_label_counts"
-                    )
                     raise RuntimeError(
                         f"Inserted {stats['inserted_records']} row(s) but "
                         f"{counts_helper} returned nothing for "
--- a/tests/test_token_classification_tag_counts.py
+++ b/tests/test_token_classification_tag_counts.py
@@ -1,0 +1,224 @@
+"""token_classification output_classes = DISTINCT TAGS, not tag SEQUENCES
+(backend#1747).
+
+The ``label`` column stores the whole per-row BIO tag sequence (one tag per
+word — e.g. ``"O B-PER I-PER O"``), so the row-unit ``get_label_counts``
+(``GROUP BY label``) counts distinct SEQUENCE STRINGS as classes — dozens of
+them — and the trained model head, built on that class set, never links to the
+9-ish real tags. The task is then unrunnable end to end.
+
+The fix routes token_classification's ingest-summary count through
+``Database.get_tag_counts``, which explodes each sequence into its tags, gated
+on the ``label_is_tag_sequence`` ModalitySpec trait (never a category string).
+
+Covered here:
+- the registry trait (set on token_classification, and unique to it);
+- the trait does NOT leak into the CLI layout contract (unchanged surface);
+- the base.py count-path routing: token_classification queries the exploded
+  helper and ships its distinct-tag counts as ``labels``; a non-tag category
+  still uses the row-unit helper;
+- ``get_tag_counts`` itself — the explode-and-weight arithmetic (the 77→9
+  bug), NULL-label handling, and SQL shape.
+
+The DB boundary is mocked like the rest of the unit suite.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, Generator
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from tracebloc_ingestor.config import Config
+from tracebloc_ingestor.database import Database
+from tracebloc_ingestor.ingestors import base as base_mod
+from tracebloc_ingestor.ingestors.base import BaseIngestor
+from tracebloc_ingestor.modalities import REGISTRY, spec_for
+from tracebloc_ingestor.modalities.layout import build_layout_contract
+from tracebloc_ingestor.utils.constants import TaskCategory
+
+TOKEN = TaskCategory.TOKEN_CLASSIFICATION
+
+
+# ---------------------------------------------------------------------------
+# Registry trait + layout contract
+# ---------------------------------------------------------------------------
+
+
+def test_token_classification_carries_tag_sequence_trait():
+    spec = spec_for(TOKEN)
+    assert spec.label_is_tag_sequence is True
+    # Sanity: it's still the file-bearing NLP text task it always was, and it is
+    # NOT a whole-sequence classifier (the per-token head has no single class).
+    assert spec.is_file_bearing and spec.is_nlp
+    assert not spec.is_classification and not spec.is_self_supervised
+
+
+def test_tag_sequence_trait_is_unique_to_token_classification():
+    # Today exactly one tag-sequence category. A second one is a deliberate
+    # registry decision — update this test when it lands.
+    tagged = {c for c, s in REGISTRY.items() if s.label_is_tag_sequence}
+    assert tagged == {TOKEN}
+
+
+def test_trait_does_not_leak_into_layout_contract():
+    # label_is_tag_sequence drives the ingest-summary count only; it is not part
+    # of the CLI's on-disk staging contract, so the layout JSON is untouched (no
+    # version bump, no new per-task key).
+    doc = build_layout_contract()
+    assert doc["version"] == "2"
+    assert "label_is_tag_sequence" not in doc["tasks"][TOKEN]
+
+
+# ---------------------------------------------------------------------------
+# base.py count-path routing (DB boundary mocked)
+# ---------------------------------------------------------------------------
+
+
+class _FakeIngestor(BaseIngestor):
+    """Concrete BaseIngestor whose read_data yields preset records."""
+
+    def __init__(self, records, **kwargs):
+        self._records = records
+        super().__init__(**kwargs)
+
+    def read_data(self, source: Any) -> Generator[Dict[str, Any], None, None]:
+        yield from self._records
+
+
+def _make_ingestor(records, **overrides):
+    db = MagicMock(name="Database")
+    # #350: content_hash is the default id strategy; the mock DB must return a
+    # real salt string (a MagicMock salt poisons the hash and drops every row).
+    db.get_or_create_table_salt.return_value = "0" * 64
+    db.create_table.return_value = MagicMock(name="table")
+    db.insert_batch.return_value = ([1, 2], [])  # ids, db_failures
+    db.get_table_schema.return_value = {"label": "VARCHAR(255)"}
+    db.get_samples.return_value = []
+    api = MagicMock(name="APIClient")
+    api.config.TITLE = None
+    api.send_ingest_summary.return_value = {"dataset_id": 1, "dataset_key": "k"}
+
+    kwargs = dict(
+        database=db,
+        api_client=api,
+        table_name="tok_tbl",
+        schema={"label": "VARCHAR(255)"},
+        label_column="label",
+        intent="train",
+        category=TOKEN,
+    )
+    kwargs.update(overrides)
+    return _FakeIngestor(records, **kwargs)
+
+
+def _run_ingest(ing, batch_size=10):
+    """Run ingest with Session / validation / file-transfer patched out so a
+    file-bearing category runs without a real filesystem."""
+    with patch.object(base_mod, "Session") as Sess, patch.object(
+        ing, "validate_data", return_value=True
+    ), patch.object(
+        base_mod,
+        "map_file_transfer",
+        side_effect=lambda c, r, o, cfg=None, source_record=None: r,
+    ):
+        Sess.return_value.__enter__.return_value = MagicMock()
+        return ing.ingest("src", batch_size=batch_size)
+
+
+def test_ingest_counts_distinct_tags_not_sequences():
+    """The whole bug: two rows carry DISTINCT BIO SEQUENCE strings but share the
+    same three TAGS. output_classes must be those tags, so the count path uses
+    the exploded helper and ships its result verbatim as ``labels`` — never the
+    row-unit or sequence-unit helpers."""
+    records = [
+        {"filename": "f1", "label": "O B-PER I-PER"},
+        {"filename": "f2", "label": "B-PER O O"},
+    ]
+    ing = _make_ingestor(records)
+    ing.database.get_tag_counts.return_value = {"O": 4, "B-PER": 2, "I-PER": 1}
+
+    failed = _run_ingest(ing)
+
+    assert failed == []
+    ing.database.get_tag_counts.assert_called_once_with("tok_tbl", ing.ingestor_id)
+    ing.database.get_label_counts.assert_not_called()
+    ing.database.get_label_sequence_counts.assert_not_called()
+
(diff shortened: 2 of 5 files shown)


tracebloc_ingestor/ingestors/base.py
```diff
@@ -1212,7 +1226,19 @@ def _ingest_with_lock(
                     self.file_options["number_of_sequences"] = sum(
                         label_counts.values()
                     )
+                elif self._label_is_tag_sequence:
```
LukasWodka, line 1229: **This dict is not only the class set — the backend derives the dataset's record count from `sum(values)`, and after this change those values are token occurrences.** Traced end to end rather than inferred:

`label_counts` → `send_ingest_summary(labels=...)` (`base.py:1337`) → backend `IngestSummaryView` puts it on `summary['labels']` (`global_meta/views.py:67`) → `create_dataset_from_summary` (`common/utils/dataset_utils.py:218`):

```python
labels = summary["labels"]
total  = sum(labels.values())          # :246
...
    "total": total,
    "unique_data_items_count": total,   # :262
```

and per-edge, in `create_dataset` (`dataset_utils.py:132`):

```python
edge_image_info = {edge: sum(edge_data["labels"].values()) ...}
```

So the trade this PR makes is narrower than the description says. **Before:** classes wrong (distinct sequences), `sum(values)` == row count, so `total` / `unique_data_items_count` were right. **After:** classes right, but `sum(values)` is the number of *tagged tokens*. A 1,000-row dataset averaging ~20 words per row now registers `unique_data_items_count = 20,000` — a field whose name is a promise about data items.

That also puts the ingestor and the backend into disagreement about the same run: the ingestor reports `stats["api_sent_records"] = stats["inserted_records"]` (rows), while the backend stores a `total` ~20× larger from the same payload.

I don't think the fix is in this file — the payload has one dict doing two jobs (class vocabulary *and* record tally), and separating them is a `send_ingest_summary` contract change plus a backend read. Which is why I'm raising it rather than suggesting a patch. What I'd want before this lands is a decision on record, one of:

- **Pair it** with the backend change so `total` / `unique_data_items_count` come from the row count for tag-sequence categories, and land them together; or
- **Ship it and track the consequence explicitly** — the class fix is clearly the more valuable half, and an inflated item count is arguably better than a task that cannot run at all. But then it belongs in the PR body and on backend#1747 as a known trade, not discovered later by someone asking why a token-classification dataset claims twenty thousand records.

The PR body already draws this boundary well for the backfill case (*"correcting a stored output_classes without a re-ingest is a backend-endpoint change, out of this repo's scope"*). This is the same boundary, on the totals, and it isn't mentioned.

One thing I checked and it's **fine**, so you don't have to: repeated tags within a sequence are counted per occurrence, not per row — `for tag in str(label).split()` visits each token, so `"O B-PER O"` with `row_count=5` correctly gives `O → 10`. The docstring's "token occurrences" is accurate.

LukasWodka (changes requested):
The diagnosis is exactly right and the mechanism is the correct one: counting `output_classes` off `GROUP BY label` when the column holds a whole BIO *sequence* is why nothing links a head, and exploding the sequence while keeping the `GROUP BY` in SQL is the fix that doesn't turn an `O(distinct sequences)` job into `O(rows)`. Reading the trait from the registry rather than branching on `self.category == "token_classification"` is the right shape too — it matches `grouping`, and it means the next tag-sequence category is a registry line.

I checked the arithmetic that the whole fix turns on, rather than trusting the docstring: repeated tags inside one sequence are counted per **occurrence**, because `for tag in str(label).split()` visits each token — `"O B-PER O"` at `row_count=5` gives `O → 10`, not 5. So "token occurrences" is accurate, and the 77→9 collapse in the test is real. The `counts_helper` refactor is also safe: all three branches of the if/elif/else assign it, so the `elif not label_counts:` message can't hit an unbound name.

**One blocker, inline, and it's outside this repo's diff — which is why it's easy to miss.** That same `labels` dict is not only the class vocabulary: the backend derives the dataset's *record count* from `sum(labels.values())`. `create_dataset_from_summary` (`backend`, `common/utils/dataset_utils.py:246`) sets both `total` and `unique_data_items_count` from it, and `create_dataset:132` computes per-edge counts the same way. Before this change those sums equalled the row count, so they were right while the classes were wrong; after it they're the tagged-token count, so a 1,000-row dataset with ~20 words a row registers `unique_data_items_count = 20,000`. The ingestor meanwhile still reports `api_sent_records = inserted_records`, so the two services disagree about the same run.

I'm not asking you to fix it here — one dict is doing two jobs (vocabulary and tally), and splitting them is a `send_ingest_summary` contract change plus a backend read. What I'd want before this merges is the decision written down: either land it with the backend side, or ship it and record the inflated count on backend#1747 as a known trade. The class fix is plainly worth more than an accurate item count — a task that can't run at all is the worse failure — but that trade should be chosen rather than discovered.

Everything else reads well, and the scoping notes in the PR body (backfill runner out of scope, CLI layout contract untouched) are the right calls.

tracebloc_ingestor/ingestors/base.py
shujaatTracebloc: Confirmed against `backend@origin/develop` rather than taken on trust — your trace holds line for line (`views.py:67`, `dataset_utils.py:246` `total = sum(labels.values())`, `:261` `unique_data_items_count`, `:131` `edge_image_info`; your `:262`/`:132` are one off, nothing else moved).

**Two things to add, one of which widens the blast radius and one of which I think points at the fix.**

**1. There is a second corrupted path you didn't name: `combine_datasets` (`dataset_utils.py:275`).** It merges labels across edges and recomputes the same quantity:

```python
for edge_data in merged_summary.values():
    for label, count in edge_data["labels"].items():
        merged_labels[label] = merged_labels.get(label, 0) + count
total = sum(merged_labels.values())        # :318
...
    "unique_data_items_count": total,      # :327
```

So the token-count semantics don't just land on the single-edge dataset — they compound through every combined one. That's the worse direction, because a combined dataset is exactly where nobody re-derives the number by hand.

**2. The same dict feeds fields whose correctness requirements point in opposite directions**, which I think is the actual shape of the problem. In the same payload:

```python
"count": {"label": labels},      # class set — the new semantics are RIGHT
"label_density": labels,         # per-tag frequency — arguably MORE right now
"total": total,                  # row count — WRONG under the new semantics
"unique_data_items_count": total # row count — WRONG, and the name is a promise
```

For BIO tags, per-tag token frequency is a *better* `label_density` than per-sequence frequency — a sequence appearing 5 times tells you nothing about how common `B-PER` is. So this isn't "the old semantics were right and the new ones are wrong". It's that one dict is being asked to answer two different questions, and #1747 only breaks the tie for one of them.

Which suggests the fix isn't choosing between token counts and row counts — it's the ingest summary carrying **both**: the per-tag tally for the class set and density, and an explicit row count for `total` / `unique_data_items_count`, rather than the backend inferring the second by summing the first. That's a backend-side change too, so it's bigger than this PR, but picking either side alone leaves one of the four fields wrong.

Not resolving this — it's your thread and your call. Recording the second path so it doesn't get missed if this gets rescoped.

shujaatTracebloc (commented):
Not approving — @LukasWodka's change-request is standing and his finding is correct. I verified it against `backend@origin/develop` rather than adding a second opinion on the same evidence, and it holds line for line. Detail on his thread; the short version is that I found a **second** corrupted path he didn't name (`combine_datasets`, `dataset_utils.py:275`), so the token-count semantics compound through combined datasets too.

**On the diagnosis itself, which I have history with:** this is the bug that made `token_classification` unrunnable end to end, and the numbers here match what I measured on dev — 77 distinct sequences where the head needs 9 distinct tags. The `GROUP BY label` over a column holding whole BIO sequences is the right root cause, and porting the explosion logic rather than restoring the deleted `_explode_tags` file is the right call: that file carried a lot more than this needs.

**Two things I'd defend if they come up in the rework:**

The **trait-style selection** (`ModalitySpec.label_is_tag_sequence`) is worth keeping even if the counting changes. Branching on a category string is how the next tag-sequence task type silently gets the wrong count path, and mirroring the existing `grouping` trait means the registry stays the one place that knows.

Keeping the `GROUP BY label` in SQL and exploding the *distinct sequences* is also right — `O(distinct sequences)` rather than `O(rows)` matters at ingest scale, and it's easy to lose that property in a rewrite that moves the explosion earlier.

**Where I'd point the rework:** the blocker isn't that the new counts are wrong, it's that one dict is answering two questions. `count.label` and `label_density` want per-tag token tallies; `total` and `unique_data_items_count` want rows. Choosing either semantics leaves two of those four fields wrong, so the summary probably needs to carry both numbers — which makes this a cross-repo change rather than an ingestor-only one.

CI is green and there are no Bugbot findings; the only thing blocking is the correctness question above.

tracebloc_ingestor/ingestors/base.py
LukasWodka: **Your corrections are right and mine were the off-by-ones.** Re-checked on `origin/develop` just now rather than defending my own numbers: `edge_image_info` is `:131`, `unique_data_items_count` in `create_dataset_from_summary` is `:261`. I cited `:132` and `:262`. `:246` we agree on. Thanks for checking rather than pasting my trace forward — that's the failure mode where a wrong citation gets laundered into fact by being repeated.

Returning the favour on one of yours: `total = sum(merged_labels.values())` is **`:313`**, not `:318` (`:327` for `unique_data_items_count` is right). The merge loop above it is `:310-312`.

**`combine_datasets` is real and it's the worse half.** Verified — `:275`, recomputing over `merged_summary` and landing on the same two fields at `:313`/`:327`. And your reason is the part that matters more than the location: a combined dataset is precisely where nobody sanity-checks the number against a row count they remember, so an inflated total there survives longer than the single-edge one. That's a second corrupted path and it should be in the record on #1747.

**Your reframing is better than mine and I'm adopting it.** I said this PR "trades one wrong number for another", and that's not accurate. Laid out per field:

| field | under the new semantics |
|---|---|
| `count.label` | **right** — the class set is the whole point of #1747 |
| `label_density` | **better** — per-tag frequency genuinely beats per-sequence frequency |
| `total` | wrong — it's a record count |
| `unique_data_items_count` | wrong, and the name is a promise |

Three of four improve or hold; one *class* of field breaks. So the choice isn't token-counts-versus-row-counts — it's that the ingest summary makes the backend infer a record count by summing a frequency map, and #1747 breaks that inference without touching it. Carrying both numbers explicitly is the fix, exactly as you say.

**Where that leaves the ask, @aptracebloc** — unchanged in substance, sharper in shape. Not "undo the tag counts": they're right. The decision on record is whether this lands with the backend side (an explicit row count in the summary payload, consumed at `:246`/`:313` instead of the `sum`), or ships now with **both** corrupted paths recorded on backend#1747 as a known consequence. I'd still take the second if the first is slow — an unrunnable task type is worse than an inflated count — but it needs to be a choice with the `combine_datasets` path named in it, not just the single-edge one.

Leaving this open; the change-request stands until that decision is written down somewhere it will outlive the PR.

aptracebloc: Decision made and written down where it outlives the PR: **ship the class fix; the inflated record-count fields are a known, accepted trade** until a cross-repo follow-up.

Recorded on **backend#1747** — https://github.com/tracebloc/backend/issues/1747#issuecomment-5406738806 — with both corrupted paths named explicitly: single-edge (`create_dataset_from_summary` `:246`/`:261`, per-edge `edge_image_info` `:131`) **and** `combine_datasets` (`:275` → `:313`/`:327`), the field-by-field table (`count.label` right, `label_density` better, `total`/`unique_data_items_count` wrong), and the follow-up shape (carry an explicit row count in the summary payload and consume it instead of `sum(labels.values())` for tag-sequence categories). The PR body now carries the short version too.

Nothing in the class fix changes — the trait selection, the `O(distinct sequences)` explode, and the per-occurrence tag counting all stand, as you both confirmed. Re-requesting review with the decision on record.

LukasWodka: Satisfied — and I checked the record rather than the link text. [backend#1747#issuecomment-5406738806](https://github.com/tracebloc/backend/issues/1747#issuecomment-5406738806) exists and carries everything the ask named: both corrupted paths with their line numbers (single-edge `:246`/`:261` plus per-edge `:131`, **and** `combine_datasets` `:275` → `:313`/`:327`), the field-by-field table, the decision stated as a choice, and the follow-up's actual shape — carry the row count explicitly in the payload and consume it instead of `sum(labels.values())` for tag-sequence categories.

It also uses the **corrected** line numbers rather than the ones I first posted, which is the detail that tells me it was written from the trace and not from my comment.

For the record, this is the option I said I'd accept, so I'm not going to re-litigate it: an unrunnable task type is the worse failure, and an inflated secondary count that is written down is a different thing from one that is discovered six months later by someone asking why a token dataset claims twenty thousand records. The distinction the decision note draws — "chosen, not discovered" — is the whole point.

The PR body's short version is good too, and it does the one thing a summary of a known trade has to do: it names `combine_datasets` as compounding, so a reader who only ever sees the PR still learns the worse path exists.

Resolving. Converting my change-request to an approval.

LukasWodka (approved):
Approving — my change-request is satisfied and my thread is resolved. Green gate clean: 14 pass, zero unresolved threads, no conflicts, and the head is unchanged (`ea44b5b`), so everything I verified on the first pass still stands.

**The class fix was never in doubt and it's right.** `GROUP BY label` over a column holding a whole BIO sequence counts sequence strings as classes; exploding the sequence while keeping the `GROUP BY` in SQL fixes the vocabulary without turning an `O(distinct sequences)` job into `O(rows)`. I checked the arithmetic the fix turns on — repeated tags inside one sequence are counted per **occurrence**, because `for tag in str(label).split()` visits each token, so `"O B-PER O"` at `row_count=5` gives `O → 10`. Reading the trait from the registry rather than branching on `self.category` is the right shape and matches `grouping`.

**What I asked for was a decision on record, and that's what landed.** The note on backend#1747 names both corrupted paths with line numbers — the single-edge one *and* `combine_datasets`, which is the one that compounds — carries the per-field table, and states the follow-up concretely enough to act on. It also uses the corrected line numbers rather than the ones I first posted, which is how you can tell it was written from the trace.

Worth recording what this review actually cost and returned, because it's a good argument for the shape:

- The finding wasn't in the diff. It was two repos away, in code this PR never opens — the backend inferring a record count by summing a frequency map.
- @shujaatTracebloc re-derived my trace instead of forwarding it, caught two off-by-one citations of mine, and found the second corrupted path I'd missed. His reframing — one dict answering two questions with opposite correctness requirements — is better than my original "trades one wrong number for another", and it's what made the follow-up's shape obvious.
- The outcome isn't a bigger PR. It's the same PR, shipping now, with a consequence that is bounded, named, and attached to the ticket that will fix it.

Nice work on all three counts.

shujaatTracebloc (approved):
Approving.

**First, my own correction, since I was the one who made a point of checking citations rather than repeating them.** @LukasWodka is right: `total = sum(merged_labels.values())` is **`:313`**, not `:318` as I wrote. Verified on `origin/develop` — `:313` for the sum, `:327` for `unique_data_items_count`, merge loop at `:310-312`. I corrected two of his line numbers and introduced one of my own in the same comment, which is exactly the laundering failure he named. The finding held; the citation didn't.

**On the decision itself, which is what I'm actually approving.** Shipping a field that's knowingly wrong is normally where I'd stop, so it's worth being explicit about why this clears:

- The alternative is worse and not hypothetical. `token_classification` is unrunnable end to end right now — no head links to a class set of 77 sequences where 9 tags are expected. An inflated secondary count is a lesser failure than a task type nobody can use.
- The trade is **written where it outlives the PR**, not just agreed in a thread. I checked the record rather than the link text: `backend#1747` comment `5406738806` exists, authored 07:03:19, and carries both corrupted paths, the field table, and the follow-up shape.
- It uses the **corrected** line numbers — including `:313`, the one I got wrong — which is the tell that it was written from the trace rather than copied from our comments.
- The PR body carries the short version, and it names `combine_datasets` as the compounding path. That matters more than it looks: a reader who never opens the issue still learns the worse path exists.

"Chosen, not discovered" is the right distinction, and it's the difference between a known trade and a bug someone finds in six months wondering why a token dataset claims twenty thousand records.

**The class fix itself I'd defend unchanged:** trait-style selection via `label_is_tag_sequence` rather than a category string, so the next tag-sequence modality is a registry entry instead of a silent miss; the `GROUP BY` staying in SQL so the explode is `O(distinct sequences)` not `O(rows)`; and porting the explosion rather than restoring the deleted `_explode_tags` file, which carried far more than this needs.

Gate: all checks green or skipped, no open threads, `MERGEABLE`, and @LukasWodka's change-request converted to an approval. Nothing outstanding from me.

PR #528, opened 2026-08-25
ci(2272): mint with client-id, not the deprecated app-id

LukasWodka:
Replaces the deprecated `app-id` input with `client-id` on every
`actions/create-github-app-token` mint in this repo (backend#2272).

The action deprecated `app-id` (`Input 'app-id' has been deprecated with message: Use
'client-id' instead.`), so every run currently prints that warning. A client id is a
**different value** from the numeric app id, so this needed a new secret before a single
line could change: **`RELEASE_TRAIN_APP_CLIENT_ID`** was created on 2026-08-25 at **both**
org Actions and org Dependabot scope, `visibility=all` — both scopes because GitHub gates
Dependabot-triggered workflows on a separate secret store. The private key
(`RELEASE_TRAIN_APP_PRIVATE_KEY`) is unchanged; only the identifier moves.

## What changed

- `.github/workflows/add-to-kanban.yml`

Exactly one line per mint:

```diff
-          app-id: ${{ secrets.RELEASE_TRAIN_APP_ID }}
+          client-id: ${{ secrets.RELEASE_TRAIN_APP_CLIENT_ID }}
```

Nothing else. In particular **no `permission-*` line is touched** — backend#2157 is
separate work on these same steps (four exemptions still stand there), and a stray scope
edit here would collide with it and muddy both.

## Ordering — this PR is **wave 1**; `tracebloc/.github` merges **LAST**

`add-to-kanban.yml` is a per-repo **copy**, and `caller-drift.py` compares each repo's
blob sha against `tracebloc/.github`'s copy as the source of truth. So every other repo
must merge **before** `.github`. This PR is in the first wave; the `.github` PR is
deliberately held until the rest have landed.

The org caller-drift audit **will report drift for the duration of the window**. That is
inherent to a byte-compared copy changing across repos and is the accepted cost — the
window is kept short by opening every PR together and merging in order.

## Verification

- `actionlint` — clean on 1 edited file(s).
- `python3 -c "import yaml,sys; yaml.safe_load(open(sys.argv[1]))"` — clean on 1 edited file(s).
- The pinned action `actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0`
  accepts `client-id` — confirmed by reading `action.yml` **at that exact SHA**, not from
  memory. All 42 mint steps across the org sit on this one pin, so no version bump is
  needed anywhere and none is made here.
- Diff is 1 file changed, 1 insertion(+), 1 deletion(-) — a pure 1:1 line swap.

### What this cannot prove

**A token mint can only be proven by a workflow actually running.** If the new secret's
value were wrong, this fails at mint time with an auth error; the first real board write
is the test. Stating that plainly rather than letting the green checks above imply
coverage they do not have.

Refs tracebloc/backend#2272

---

> [!NOTE]
> **Low Risk**
> Low risk: one credential identifier change on an existing mint step; wrong secret value fails at mint time rather than silently widening scope.
> 
> **Overview**
> Updates the **add-to-kanban** workflow so `actions/create-github-app-token` uses the **`client-id`** input and **`RELEASE_TRAIN_APP_CLIENT_ID`** instead of the deprecated **`app-id`** / **`RELEASE_TRAIN_APP_ID`**.
> 
> That is a single-line swap on the mint step; **`private-key`**, **`owner`**, **`repositories`**, and all **`permission-*`** lines are unchanged. The new secret must exist in both org Actions and Dependabot scopes so Dependabot-triggered runs still mint a token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9acb4848765d34bcb622507f00d34260f3358799. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/add-to-kanban.yml
+++ b/.github/workflows/add-to-kanban.yml
@@ -27,7 +27,7 @@ jobs:
         id: app-token
         uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         with:
-          app-id: ${{ secrets.RELEASE_TRAIN_APP_ID }}
+          client-id: ${{ secrets.RELEASE_TRAIN_APP_CLIENT_ID }}
           private-key: ${{ secrets.RELEASE_TRAIN_APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
           # SCOPED TO THIS REPO, or the two reads below land org-wide (saadqbal,


shujaatTracebloc (changes requested):
Requesting changes: this one-line swap breaks the workflow it edits, and it does so on **every repo in the rollout**, not just here.

## It's the change, not pre-existing red

`add-to-kanban.yml` has succeeded on every run for the past week — release-train promotions, docs PRs, other `ci/` branches. The only failing run is `ci/2272-client-id-mint` itself:

```
2026-08-25  failure  ci/2272-client-id-mint
2026-08-24  success  release-train/to-main
2026-08-24  success  release-train/to-staging
2026-08-24  success  docs/2298-deprecate-tf-readme
2026-08-22  success  ci/1979-stale-backlog-caller
...
```

## The failure

Both red checks die at the same step, `Mint an installation token`:

```
POST /repos/tracebloc/<repo>/installation → 401
  "message": "'Issuer' claim ('iss') must be an Integer"
```

## Scope: all seven instances, identically

I checked each one rather than assuming the rollout was uniform:

| PR | failing |
|---|---|
| `start-training#72` | `add-to-project`, `set-status` |
| `model-zoo#178` | `add-to-project`, `set-status` |
| `data-ingestors#528` | `add-to-project`, `set-status` |
| `tracebloc-py-package#448` | `add-to-project`, `set-status` |
| `backend#2484` | `add-to-project`, `set-status` |
| `averaging-service#378` | `add-to-project`, `set-status` |
| `tracebloc-engine#727` | same, plus an unrelated `test (3.11)` |

Seven for seven. Whatever the cause, it isn't repo-local configuration drift.

## What the error narrows it to

Worth separating what's measured from what's inferred, because the fix differs.

**Measured:** the action got as far as signing a JWT and calling the API. That matters — if `RELEASE_TRAIN_APP_CLIENT_ID` were unset, `create-github-app-token` would refuse before minting anything, for want of either identifier. So the secret almost certainly holds a value; GitHub rejected the value it holds.

**Inferred:** GitHub is being told `iss` must be an integer, which is what it says when the issuer isn't an app ID. Two candidates, and they're cheap to tell apart:

1. **The secret holds something that isn't a client ID** — a stale app ID copied into the new secret name, or a truncated paste. Check its shape: a client ID looks like `Iv23li…`, an app ID is all digits.
2. **The pinned action version doesn't map `client-id` onto the JWT the way this assumes.** `create-github-app-token` is pinned at `bcd2ba49…` (v3.2.0); worth confirming that version signs with the client ID as issuer rather than still expecting an app ID.

## The good part

These PRs edit the very workflows that run on them, so the change tests itself on the way in. That's why this surfaced as seven red PRs instead of seven repos whose board automation quietly stopped working after merge — the kanban jobs are exactly the kind that fail unnoticed, because nobody watches a card that didn't move.

Happy to re-review as soon as the mint succeeds. If it turns out the secret is right and `client-id` genuinely isn't supported at this action version, the honest outcome may be that `app-id` isn't deprecated for this path yet — worth saying so on backend#2272 rather than routing around it.

LukasWodka:
## ⛔ Converted to draft — the `client-id` mint fails 401. Do not merge.

The change is mechanically correct, but **it does not work**, and the first workflow run
proved it. Flagging in the PR body that "a token mint can only be proven by a workflow
actually running" was not hypothetical — it ran, and it failed.

### Evidence

`add-to-project` on this PR (the job that runs the changed `add-to-kanban.yml`):

```
Run actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1
  client-id: ***
  private-key: ***
  owner: tracebloc
Failed to create token for "tracebloc/backend" (attempt 1):
  'Issuer' claim ('iss') must be an Integer - https://docs.github.com/rest
##[error]'Issuer' claim ('iss') must be an Integer
  status: 401
  url: 'https://api.github.com/repos/tracebloc/backend/installation'
```

The action signs a JWT using the supplied identifier as the `iss` claim. GitHub rejected
that JWT with **401**, so no installation token is minted and the board write never happens.

### It is one root cause, not two

`set-status` is also red on these PRs, but **not** for its own reason — it still mints
through `.github@main`, which is untouched and mints fine (its log still prints the
`app-id` deprecation warning). It failed *downstream*:

```
##[error]PR #NNNN is not on project 2 after 5 retries. add-to-kanban should have
added it; it did not, or it lost the race.
```

That is `set-status` failing closed correctly because the card was never created. Fix the
mint and both go green.

### What is ruled out

- **Not the action version.** `v3.2.0` (`bcd2ba49…`) is the **latest** release, and its
  `action.yml` at that exact SHA declares `client-id` as a supported input. There is
  nothing to bump to, and no pin was bumped here.
- **Not a syntax or scoping error.** `actionlint` and a YAML parse are clean on every
  edited file; the diff is a pure 1:1 line swap; `mint-scope.py` still finds all 17 mints
  with 0 findings, so no `permission-*` input moved.
- **Not pre-existing.** `add-to-project` and `set-status` are `SUCCESS` on PRs that predate
  this branch.

### What this points at

`RELEASE_TRAIN_APP_CLIENT_ID` needs its **value** checked. GitHub is reporting that the
identifier it received is not one it will accept as a JWT issuer for this App. That is a
secret-value question, not a code question, and it needs someone who can read the secret —
the App's client ID as shown on its settings page (the `Iv23li…` string), with no
surrounding whitespace or newline.

Once the value is confirmed, take this PR out of draft and re-run; nothing in the diff
needs to change if the secret was simply wrong.

Refs tracebloc/backend#2272

LukasWodka:
Closing as **blocked upstream**, not abandoned — see backend#2272 for the full evidence.

The migration is written and validated, and the mint fails at runtime:

```
Failed to create token: 'Issuer' claim ('iss') must be an Integer   status: 401
```

That error proves the action worked: it *did* send the client ID as the JWT `iss` claim, and **GitHub rejected it**. Everything on our side was eliminated first — the pin is genuinely `v3.2.0` (the latest release) and its `action.yml` declares `client-id`; the diff is a pure 1:1 line swap with `actionlint` and YAML parse clean; the stored secret is the App's Client ID; and a re-set with `read -rs` + `printf '%s'` (no trailing byte) failed identically.

**`app-id` keeps working — the deprecation is a warning, not a breakage** — so nothing degrades by staying put.

Not lost: 20 files, 37 lines, a pure swap, and this PR's diff remains readable here. Re-open when GitHub accepts client IDs as issuers for this App.

PR #536, opened 2026-08-27
ci: refuse a schema-version bump our consumers cannot read (backend#2713)

LukasWodka:
## The gap

This repo publishes versioned contract files. `tracebloc/e2e-test-agent` reads them and accepts an **explicit** set of versions, refusing an unknown one by design. **Nothing checked that the two agree.**

On 2026-08-27 `layout.v1.json` went 2 → 3 (#535). It landed green here and turned the **required** `unit` context red on **every open PR** in that repo, plus its journey, for **68 minutes**. Its guard caught it at *run* time; nothing caught it at *merge* time.

#290 over there fixed the instance — read the diff, added `"3"`, and took the ordering column the new contract declares. This closes the class.

## Both sides derived, neither restated

**There is no table in this PR pairing a contract with a consumer, and there must never be one** — a hand-written pairing agrees with itself while disagreeing with reality, which is the defect this exists for.

Every consumer module already declares *both* halves itself:

```python
CONTRACT_RELPATH  = Path("tracebloc_ingestor") / "schema" / "layout.v1.json"   # which file
SUPPORTED_VERSIONS = {"2", "3"}                                                # which versions
```

So the pairing is **discovered** by walking the consumer for modules carrying both, and the producer side is the `version` in the file that module names. Add a third contract on either side and this picks it up with no edit here. Today it finds two, unprompted:

```
harness/layout.py       reads tracebloc_ingestor/schema/layout.v1.json      and accepts ['2', '3']
harness/runtime_env.py  reads tracebloc_ingestor/schema/runtime_env.v1.json and accepts ['1']
```

Read from the **AST, not a regex**: a commented-out or runtime-built `SUPPORTED_VERSIONS` must not count as a declaration, and a substring search accepts both while reporting agreement with a module that declares nothing.

## Fails closed, including on "cannot tell"

**Zero discovered pairs is an error, not a pass.** Zero parsed pairs compares equal to zero parsed pairs, so shrugging there buys a check that passes for ever the day the consumer's layout moves — which is this very defect class. Same for an unreadable consumer, a contract the consumer names that we do not publish, and a contract with no `version` field.

## The public/private asymmetry, handled rather than papered over

This repo is public and the consumer is private, so the job needs an App token — and GitHub withholds secrets from fork PRs. That means **the job cannot be a required context**, and a guard whose only exercise is a non-required job is advice, not a gate.

So the work is split:

- **`tests/test_contract_consumers.py` runs inside the required `pytest` context** and covers the comparison logic against fixtures. That is what makes the mechanism real.
- **The job supplies the real inputs** — the actual two repos.

The fork case is decided by what the PR changed, not by a blanket skip:

| | no token available |
|---|---|
| no schema file in the diff | **notice + pass** — there is nothing this check could verify |
| a schema file in the diff | **fail** — an external contributor cannot know the private consumer's state, and a contract bump is exactly the change needing an internal look |

A nightly backstop covers the other direction: the consumer can *narrow* its supported set in its own repo, which no PR here would ever notice.

## Evidence

**Replayed the real incident.** With the consumer pinned at the commit before its fix (`SUPPORTED_VERSIONS = {"2"}`) and this repo publishing `3`:

```
tracebloc_ingestor/schema/layout.v1.json: this repo publishes version '3',
    but harness/layout.py accepts only ['2'].
    The consumer refuses an unknown version BY DESIGN, so merging this bump
    turns its required checks red on every open PR until it is taught to read '3'.
    Fix order: land the consumer change FIRST, then merge this.
exit 1
```

Against today's two repos it exits 0. **A check written after an incident that would not have caught the incident is decoration**, so both directions are asserted in the suite, not just demonstrated here.

`2179 tests pass` (17 new), `ruff` clean, every action pinned to a SHA already in use in this repo.

## Reviewer notes

- The message names the **fix order** (consumer first). Someone who does not know that will merge these in the order that causes the outage, so the check says it rather than assuming it.
- `.consumer/` is a CI-only checkout path; nothing in the package or the published artifact changes.
- No runtime code is touched — this is CI plus a script plus tests.

Closes tracebloc/backend#2713

Refs backend#2695 (the instance) · #535 · tracebloc/e2e-test-agent#290

<sub>Built with Claude Code (Opus 5).</sub>

---

> [!NOTE]
> **Medium Risk**
> Changes CI and cross-repo contract enforcement with GitHub App tokens, but does not alter runtime ingestor code; a misconfigured workflow could still block merges or miss drift if tests rot.
> 
> **Overview**
> Adds a **merge-time gate** so publishing a bumped schema version here cannot silently break `tracebloc/e2e-test-agent`, which only accepts explicitly declared contract versions.
> 
> **`scripts/contract_consumers.py`** discovers each consumer module’s `CONTRACT_RELPATH` and `SUPPORTED_VERSIONS` via AST (not regex), reads the producer JSON `version`, and **fails closed** on unreadable consumers, missing files, or zero discovered pairs. Disagreements print the **fix order** (update the consumer first).
> 
> **`.github/workflows/contract-consumers.yml`** runs that script against a private consumer checkout (App token, `develop` on both sides for scheduled runs). Fork PRs without secrets **pass only when the PR does not touch schemas**; schema changes without a token **fail**. Scope uses the PR files API (not `git diff`) and avoids pipefail/SIGPIPE pitfalls.
> 
> **`tests/test_contract_consumers.py`** exercises discovery, fail-closed paths, the real `layout.v1.json` 2→3 incident replay, and pins workflow safety properties (token scope, `persist-credentials: false`, producer ref pinning).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99b4a0a04767e4b5deb555d322ca056d7a8e48a7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/contract-consumers.yml
+++ b/.github/workflows/contract-consumers.yml
@@ -1,0 +1,139 @@
+name: contract consumers
+
+# Refuse a schema-version bump the consuming repo cannot read.
+#
+# WHY (backend#2695). On 2026-08-27 `layout.v1.json` went 2 -> 3 (#535). It
+# landed green here and turned the REQUIRED `unit` context red on every open PR
+# in `tracebloc/e2e-test-agent` -- which reads that file and accepts an explicit
+# version set -- for 68 minutes. The consumer's guard caught it at RUN time;
+# nothing caught it at MERGE time, because nothing connected the two repos.
+#
+# The comparison lives in `scripts/contract_consumers.py` and its LOGIC is
+# covered by `tests/test_contract_consumers.py`, which runs inside the REQUIRED
+# `pytest` context. That split is deliberate: this job needs a token and a second
+# checkout, so it cannot be required on a fork PR, and a check whose only
+# exercise is an unrequired job is advice rather than a gate (CLAUDE.md rule 2).
+# CI supplies the real inputs; the required suite proves the comparison works.
+#
+# THE PUBLIC/PRIVATE ASYMMETRY, AND WHY THE SKIP IS NARROW. This repo is PUBLIC
+# and the consumer is PRIVATE, so reading it needs an App token -- and GitHub
+# withholds secrets from fork PRs. Rather than fail every external contribution,
+# the job asks what the PR actually changed:
+#
+#   * no schema file touched  -> there is nothing this check could tell us, so a
+#     missing token is a NOTICE and the job passes. Not a hole: the risk it
+#     guards does not exist on such a PR.
+#   * a schema file touched, no token -> FAILS. An external contributor cannot
+#     know the private consumer's state, and a contract bump is exactly the
+#     change that needs an internal look. "Cannot tell" must not read as "fine".
+
+on:
+  pull_request:
+    paths:
+      # Only PRs that can actually break a consumer. A version cannot change in
+      # a file that is not in the diff, and running this on every PR would spend
+      # a token mint to re-answer a question nothing changed.
+      - 'tracebloc_ingestor/schema/**'
+      - 'scripts/contract_consumers.py'
+      - '.github/workflows/contract-consumers.yml'
+  push:
+    branches: [develop]
+    paths:
+      - 'tracebloc_ingestor/schema/**'
+  # A BACKSTOP, because the consumer can drift without us: it may NARROW its
+  # supported set in its own repo, which no PR here would ever notice.
+  schedule:
+    - cron: '37 5 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: contract-consumers-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  agreement:
+    name: consumers can read what we publish
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: '3.11'
+
+      # Does this run actually touch a contract? Drives whether an absent token
+      # is tolerable, so it is answered from the diff rather than assumed.
+      - name: Decide whether a contract is in play
+        id: scope
+        env:
+          EVENT: ${{ github.event_name }}
+          BASE: ${{ github.event.pull_request.base.sha }}
+          HEAD: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT" != "pull_request" ]; then
+            # A scheduled or pushed run has no diff to consult and exists to
+            # check the current state, so it always needs the real comparison.
+            echo "schema-touched=true" >> "$GITHUB_OUTPUT"
+            echo "not a pull request — treating the contract as in play"
+            exit 0
+          fi
+          git fetch --no-tags --depth=1 origin "$BASE" 2>/dev/null || true
+          if git diff --name-only "$BASE" "$HEAD" 2>/dev/null \
+               | grep -q '^tracebloc_ingestor/schema/'; then
+            echo "schema-touched=true" >> "$GITHUB_OUTPUT"
+            echo "this PR changes a published contract"
+          else
+            echo "schema-touched=false" >> "$GITHUB_OUTPUT"
+            echo "no published contract in this diff"
+          fi
+
+      - name: Mint a token for the private consumer
+        id: token
+        # `continue-on-error` so an ABSENT secret (a fork PR) reaches the step
+        # below, which decides what that means. Without it the job dies here
+        # with "missing input", which reads as a broken workflow rather than as
+        # the deliberate refusal it should be on a schema PR.
+        continue-on-error: true
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.RELEASE_TRAIN_APP_ID }}
+          private-key: ${{ secrets.RELEASE_TRAIN_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: e2e-test-agent
+
+      - name: Refuse to guess when a contract changed and the consumer is unreadable
+        if: steps.token.outcome != 'success'
+        env:
+          SCHEMA_TOUCHED: ${{ steps.scope.outputs.schema-touched }}
+        run: |
+          set -euo pipefail
+          if [ "$SCHEMA_TOUCHED" = "true" ]; then
+            echo "::error::This PR changes a published contract, but the consumer could not be read (no token — forks do not receive secrets)."
+            echo "::error::Refusing to report agreement this run could not verify: a version bump the consumer cannot read turns its REQUIRED checks red on every open PR."
+            echo "::error::A maintainer must run this from a branch in this repo, where the token is available."
+            exit 1
+          fi
+          echo "::notice::No published contract in this diff and no consumer token available — nothing this check could verify, so it is not claiming anything."
+
+      - name: Check out the consumer
+        if: steps.token.outcome == 'success'
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: tracebloc/e2e-test-agent
+          # `develop` NAMED, not the repo default: that is the ref its own CI
+          # reads this repo at, so this compares the two heads that actually
+          # meet. A default-branch read here would answer a question nobody asked.
+          ref: develop
+          token: ${{ steps.token.outputs.token }}
+          path: .consumer
+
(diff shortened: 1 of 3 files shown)


LukasWodka:
Fixed in `aa9f9e8`, and this one was worth catching loudly.

The scope step decided whether a missing token was tolerable, and it inferred the diff with `git diff "$BASE" "$HEAD"` against a **default checkout** — a depth-1 merge commit, so the head SHA is frequently not in the local repo at all. The diff failed, `2>/dev/null` swallowed the error, and `schema-touched` fell through to `false`. A fork schema bump, or any token-mint failure, would then have taken the notice-and-pass branch instead of refusing.

That is **"could not tell" reading as "nothing changed"** — the precise defect this workflow exists to prevent, sitting in its own scope step, on a path nothing could see.

**Now asked of the API** (`pulls/N/files`) rather than inferred from git, so a shallow checkout cannot silently answer wrong. `github.token` is present on fork PRs too, so it works everywhere the check runs; `pull-requests: read` added.

**Every branch that cannot answer now chooses "a contract might have changed"** — unusable PR number, failed API call, empty file list. Exactly one path writes `false`, and it is the one that positively established no contract changed.

Three tests pin it, derived from the step rather than restating it:

- no `git diff` in the step's **code** — comments are stripped first, so the comment explaining the fix cannot satisfy its own guard
- exactly **one** `schema-touched=false` write; a second is a new way to fail open
- the refusal branch still exits non-zero

Both mutations redden their own guard and nothing else: reinstating the git inference, and flipping a cannot-tell branch to `false`.

2182 tests pass, ruff clean, and `consumers can read what we publish` is still green against the real consumer.

LukasWodka:
All three fixed in `3113cee`, and sweeping for the first one's **class** found a fourth instance this review had not flagged — which is the argument for pinning classes rather than reported lines.

**1 · SIGPIPE made the scope step answer `false`.** `printf | grep -q` under `set -o pipefail`: grep exits on the first match, printf takes SIGPIPE, the pipeline returns 141, and the `if` reads that as *no match*. In a fail-closed gate that is the worst possible direction. The list is written to a file and grepped there — the pipe is removed rather than reasoned about.

The fourth instance was the PR-number validation, same shape. It happens to fail *closed* because `if !` inverts the 141 — luck, not design, and not a property to leave in a gate like this. It's a `case` now.

**2 · The nightly compared the wrong two heads.** A `schedule` or `workflow_dispatch` run checks out this repo's **default** branch while the consumer is pinned at `develop`, so the backstop compared a released snapshot against the head that actually reads us. A support-set narrowing on `develop` could stay green until promotion; dropping an older released version could fail while `develop` agreed. The producer checkout now pins `develop` for those events and keeps `github.ref` for PR and push.

**3 · The token was over-scoped and left on disk.** Correct on both halves — and the second is the sharper one: `actions/checkout` stored it in `.consumer/.git`, after which this job runs **the PR's own script** against that checkout. Now `permission-contents: read` plus `persist-credentials: false`.

Four tests pin the classes, each mutation-proved. One detail worth stating: the SIGPIPE mutation **asserts it actually applied** before its result is believed — my first attempt at it had a quoting error, the mutation silently didn't happen, and the suite reported 24 passed. An inert mutation and a working guard produce identical logs, so the anchor is verified rather than assumed.

2186 tests pass, ruff clean, and the check still exits 1 on the replayed incident and 0 on today's two repos.

shujaatTracebloc (commented):
This is the right shape for the problem, and I have a specific stake in saying so — the 68 minutes you're describing came from a PR I reviewed and approved. I checked #535 carefully inside this repo, including re-deriving the new `ordering` block against the validators that enforce it, and never asked who reads the `version` field. A gate that asks that question at merge time is exactly what was missing.

Two design choices are the ones I'd defend if anyone tries to simplify them later:

**Refusing a pairing table.** *"A hand-written pairing agrees with itself while disagreeing with reality"* is precisely the failure. Discovering the pairing by walking the consumer for modules that declare both halves means a third contract is covered with no edit here — and, more importantly, it can't quietly stop covering one. A table would have been half the code and would have rotted the first time someone added a contract without updating it.

**AST, not regex.** A commented-out or runtime-built `SUPPORTED_VERSIONS` reporting agreement is exactly the vacuous-guard shape that's bitten this org repeatedly today. Reading the declaration rather than the text is what makes "this consumer accepts these versions" a fact rather than a substring.

### The residual, with current numbers

The discovery is scoped to `e2e-test-agent` and keyed on the `CONTRACT_RELPATH` + `SUPPORTED_VERSIONS` shape. That's a defensible boundary, but `layout.v1.json` has consumers outside both. I measured these just now:

```
data-ingestors (source)                                    version=3
cli/internal/schema/layout.v1.json                         version=2
backend/global_meta/tests/contracts/.../layout.v1.json     version=2
```

Both are **vendored byte-copies**, so they declare no `SUPPORTED_VERSIONS` and this walk cannot see them. They've been stale for about seven hours.

The reason I'd raise it rather than file it away: the two failure modes are opposite, and the quiet one is worse. `e2e-test-agent` pinned versions explicitly and failed closed — which is why it went red in 68 minutes and got fixed the same day. A vendored copy just drifts. Nobody has noticed these two, and nothing will make them notice.

The backend one is worth a specific look: it's a *contract test fixture*. A stale fixture means that test is green against a contract that no longer exists, which is a false green rather than a missing check — I haven't read the test, so I'd rather flag it than assert it. `cli#413` is already open as a canary for the CLI copy drifting from data-ingestors HEAD, so that half may be covered.

None of that blocks this. Closing the loud class first is the right order, and a follow-up that walks vendored copies for byte-equality against the source is a different check with a different trigger. Worth naming in the issue so the second half doesn't get assumed-done because the first half landed.

Checks are still pending at this head, so I'll pick it up again when they report — nothing blocking from me.

LukasWodka:
bugbot run

LukasWodka:
Thank you for this — and for the stake you named. Bugbot's finding on the walk is fixed in `99b4a0a` (annotated declarations were invisible; the zero-pairs guard can't see a *partial* miss, so a still-parseable sibling kept it green while the module gating the bumped contract went unread).

On the residual: **your three numbers are exactly right**, verified independently just now — source `3`, `cli/internal/schema/layout.v1.json` `2`, `backend/global_meta/tests/contracts/data_ingestors/layout.v1.json` `2`.

**But the conclusion needs correcting, and in your favour.** Those two are not drifting unnoticed — they are *deliberately pinned*, with a mechanism:

- `backend/scripts/.data-ingestors-ref` pins `470bf2a5` (data-ingestors master HEAD as of 2026-07-23) and says why in the file: *"CI's schema-drift job checks against THIS ref, not a floating branch, so an unrelated upstream commit can't red backend PRs."*
- `backend/.github/workflows/schema-drift.yml` runs `scripts/sync-ingest-schema.sh --check`.
- `cli` mirrors the pattern with its own `scripts/.data-ingestors-ref` (`8f89aece`) and a `backend-fixtures-drift.yml`.
- The adoption procedure is documented in `test_layout_contract.py`'s docstring: bump the SHA, re-sync, update `FILE_BEARING_CATEGORIES`, commit as one deliberate PR.

So the backend fixture is **not a false green**. It is green against the contract it *declares* it pins, and lagging is the design — the lag is ~5 weeks, not 7 hours, and that is a considered adoption cadence rather than rot. I'd have flagged it the same way you did without reading the pin, so this is confirmation that flagging beat asserting.

**What is a real gap, and narrower than either of us said:** `schema-drift` is **not in `backend`'s required contexts on `develop`**. So the mechanism catches drift but cannot block it — CLAUDE.md rule 2, a guard in a non-required job is advice. Same question worth asking of `cli`'s equivalent.

I've recorded all of this in backend#2713 so the second half isn't assumed-done because the first half landed — including that the vendored-copy class is **out of scope here and still open**, exactly as you asked.

shujaatTracebloc (approved):
All 14 checks green at `99b4a0a0`, no unresolved threads. Approving.

This is the same head I read last pass, so my verification stands: the `AnnAssign` fix handles both node types, applies to the shared `targets` list so it covers `CONTRACT_RELPATH` as well as `SUPPORTED_VERSIONS`, and guards on `node.value is not None` — a bare annotation with no assignment correctly does *not* count as a declaration. That last detail is the one worth keeping, because it extends the same "must not count" rule the docstring already applies to commented-out declarations, and it's exactly the case a looser fix would have let through.

What I'd want remembered about this PR, beyond the diff: the two structural choices are what make it durable. Refusing a hand-written pairing table — *"a hand-written pairing agrees with itself while disagreeing with reality"* — means a third contract is covered without an edit here, and more importantly can't quietly stop being covered. And reading declarations from the AST rather than the text is what makes "this consumer accepts these versions" a fact rather than a substring match. Both are the kind of thing a later simplification would strip out as over-engineering; they aren't.

My one open note stays a follow-up, not a blocker: the walk is scoped to `e2e-test-agent` and to modules declaring the two names, so the vendored byte-copies of `layout.v1.json` in `cli` and `backend` remain invisible to it — both still at v2 against a source at v3. That's the quiet half of the same problem, with the opposite failure mode: the version pin failed closed and got fixed inside 68 minutes, while a stale vendored copy just sits there. Worth naming in backend#2713 so the second half doesn't get treated as done because this landed.

Nice piece of work — this is the first of today's three contract-drift incidents to get a fix that generalises rather than patching the instance.

PR #537, opened 2026-08-27
release-train: develop -> staging

LukasWodka:
Automated promotion by the release train (RFC-0008 D14). Head is the train-managed `release-train/to-staging` branch (a mirror of `develop`), so it never collides with a human PR. Merged only when the fr-gate is green. 

---

> [!NOTE]
> **Low Risk**
> Contract and test surface changes only; ingest validation behavior is unchanged. Consumers must adopt layout v3’s new `ordering` field if they mirror the JSON.
> 
> **Overview**
> The layout contract (**version 3**) now includes a derived per-task **`ordering`** block so CLI and other consumers can see which tasks require monotonic time columns and whether that applies **dataset-wide** or **per group**, instead of inferring it only from validator behavior (backend#1870).
> 
> **`ordering`** is built from **`FIXED_TIME_COLUMN_BY_CATEGORY`** (column) and the spec’s **`grouping`** trait (scope). Time series forecasting gets `timestamp` / `dataset`; time series classification gets `timestamp` / `per_group`; most tasks and **time_to_event_prediction** emit explicit **`null`** (duration column, not a monotonic ordering axis). **`schema/layout.v1.json`** is regenerated; package version bumps to **0.8.14**.
> 
> Tests pin registry derivation, spot-check the three time-series cases, and **anchor contract `scope` to the validator types** each category’s factory actually runs so grouping-derived scope cannot drift from enforcement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed2d051ea335e2c4f2940268e49150a60af19b23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_layout_contract.py
+++ b/tests/test_layout_contract.py
@@ -19,7 +19,10 @@ from tracebloc_ingestor.modalities.layout import (
     contract_path,
     render_contract,
 )
-from tracebloc_ingestor.modalities.registry import REGISTRY
+from tracebloc_ingestor.modalities.registry import (
+    FIXED_TIME_COLUMN_BY_CATEGORY,
+    REGISTRY,
+)
 from tracebloc_ingestor.utils.constants import TaskCategory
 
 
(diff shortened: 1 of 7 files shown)


LukasWodka:
bugbot run

PR #538, opened 2026-08-27
fix(devex): pre-push hook parent-escape + toolchain hard-block

aptracebloc:
## What

Two defects in the fleet-wide devex pre-push hook (added in backend#1606), fixed together so the hook fleet stays consistent.

**1. `core.hooksPath` parent-escape.** `install-hooks` canonicalized `dirname` of the resolved hooks path, so `core.hooksPath=..` (or `./..`) kept the dirname inside the worktree and `install-hooks` wrote `pre-push` into the shared **parent** directory — the exact shared-dir stomp the guard exists to prevent. It now canonicalizes the resolved hooks dir itself (walks to the deepest existing ancestor + `pwd -P`, so `..` resolves), and treats the git-common-dir as in-repo so linked worktrees still install.

**2. Toolchain hard-block.** The hook soft-exited only when `make` was missing, then `exec make check` hard-failed when the venv toolchain (ruff/pytest) was absent. GUI/IDE git clients (VS Code, Tower, GitKraken) commonly have system `make` on a thin PATH but not the virtualenv, and many expose no `--no-verify` — so pushes were blocked exactly where the surrounding comments promise graceful degradation. A new `guard-toolchain` target probes the actual tools; the hook runs `make guard-toolchain` and soft-passes when it fails. Mirrors the pattern backend#1995 shipped to the JS toolchain repos (design-system, docs, frontend-app).

## Tests

`make test-hooks` — adds a hook test suite covering `core.hooksPath=..` (nothing written outside the worktree) and a missing-toolchain push (soft-pass, `make check` not run), plus a direct `guard-toolchain` pass/fail check. Green locally.

The `install-hooks` guard block is now byte-identical across the hook fleet.

Part of the fleet-wide fix for tracebloc/backend#1749.

— drafted with Claude Code

---

> [!NOTE]
> **Low Risk**
> Changes are limited to Makefile hook installation and local pre-push behavior; no runtime ingestion or CI test gates are altered beyond optional `make test-hooks`.
> 
> **Overview**
> Fixes two pre-push / `install-hooks` defects and adds **`make test-hooks`** coverage.
> 
> **Toolchain hard-block:** Adds **`guard-toolchain`** (probes that `$(PYTHON)` can run **ruff** and **pytest**). **`make check`** depends on it. The generated pre-push hook runs **`make guard-toolchain`** (with **`make -n`** first so branches without that target still run **`make check`**) and **soft-passes** when the venv tools are missing—matching the existing “missing `make`” skip for thin-PATH GUI git clients.
> 
> **`core.hooksPath` parent-escape:** **`install-hooks`** no longer trusts **`dirname`** of the hooks path alone. It canonicalizes the resolved hooks directory (ancestor walk + **`pwd -P`**), treats **`git-common-dir`** as in-repo for linked worktrees, and refuses suffix paths containing **`..`** so values like **`..`**, **`./..`**, or **`missing/../..`** cannot write **`pre-push`** outside the worktree. In-repo paths (including not-yet-created dirs like **`freshhooks`**) still install.
> 
> **Tests:** **`scripts/tests/test-pre-push-hook.sh`** exercises install idempotency, foreign-hook preservation, hooksPath inside/outside/escape cases, hook skip/soft-pass behavior, and **`guard-toolchain`** pass/fail—invoked via **`make test-hooks`**, not **`make check`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21aab12eae4bcff387d0ddc5c5d90c20e4db53ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,11 +183,14 @@ install-hooks:
 	elif hp="$$(git config --get core.hooksPath 2>/dev/null || true)"; [ -n "$$hp" ] && { \
 	       hd="$$(git rev-parse --git-path hooks)"; \
 	       case "$$hd" in /*) hdd="$$hd";; *) hdd="$$PWD/$$hd";; esac; \
-	       hdx="$$hdd"; \
+	       hdx="$$hdd"; tail=''; \
 	       while [ ! -d "$$hdx" ] && [ "$$hdx" != "$$(dirname "$$hdx")" ]; do \
-	         hdx="$$(dirname "$$hdx")"; \
+	         tail="$$(basename "$$hdx")/$$tail"; hdx="$$(dirname "$$hdx")"; \
 	       done; \
 	       chd="$$(cd "$$hdx" 2>/dev/null && pwd -P || true)"; \
+	       oIFS="$$IFS"; IFS=/; for seg in $$tail; do \
+	         case "$$seg" in ''|.) : ;; ..) chd="$${chd%/*}"; [ -n "$$chd" ] || chd=/ ;; *) chd="$$chd/$$seg" ;; esac; \
+	       done; IFS="$$oIFS"; \
 	       ctop="$$(cd "$$(git rev-parse --show-toplevel)" && pwd -P)"; \
 	       cgd="$$(cd "$$(git rev-parse --git-common-dir)" 2>/dev/null && pwd -P || true)"; \
 	       inr=0; \
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,32 @@ help:
 # running and finishes in ~11 s (measured, macOS). The coverage gate is
 # the only thing held back, and only because a floor is a merge concern
 # rather than a pre-push one.
+# guard-toolchain: can this shell actually run `make check`? A GUI/IDE git client
+# (VS Code, Tower, GitKraken) launches the pre-push hook on a thin PATH that has
+# system `make` and often a bare `python3`, but NOT the virtualenv these checks
+# were installed into — so `make check` hard-fails on "No module named ruff" with
+# no --no-verify to escape (backend#1749; the make-vs-toolchain shape of
+# backend#1995). `check` depends on this and the hook runs it first, so a push
+# from such a client degrades to a skip instead of a hard block.
+#
+# It asks the TOOLS, not the DEPENDENCIES: whether $(PYTHON) can run the linters
+# and test runner check invokes, not whether every package is installed. A
+# genuinely broken venv still fails `make check` when run from a real shell.
+.PHONY: guard-toolchain
+guard-toolchain:
+	@command -v "$(PYTHON)" >/dev/null 2>&1 || { \
+	  echo "not on PATH: $(PYTHON) — activate your virtualenv, then run: make setup"; \
+	  exit 1; }
+	@missing=''; \
+	for m in ruff pytest; do \
+	  "$(PYTHON)" -m "$$m" --version >/dev/null 2>&1 || missing="$$missing $$m"; \
+	done; \
+	[ -z "$$missing" ] || { \
+	  echo "$(PYTHON) cannot run:$$missing — activate your virtualenv, then run: make setup"; \
+	  exit 1; }
+
 .PHONY: check
-check: lint test
+check: guard-toolchain lint test
 	@echo "==> check: green (the coverage gate is in 'make check-all')"
 
 .PHONY: check-all
@@ -144,6 +168,14 @@ e2e:
 # `git rev-parse --git-path hooks` (not a hard-coded `.git/hooks`) so it lands
 # in the right place inside a linked worktree or a submodule, where the git dir
 # is not `.git`.
+# test-hooks: run the install-hooks / pre-push-hook behaviour suite on demand.
+# Not a `check` dependency: it runs real `make` in throwaway repos, so an
+# environment quirk (old git, noexec /tmp) could block a local push on something
+# CI never sees. Run it directly with `make test-hooks` (backend#1749).
+.PHONY: test-hooks
+test-hooks:
+	@sh scripts/tests/test-pre-push-hook.sh
+
 .PHONY: install-hooks
 install-hooks:
 	@if ! git rev-parse --git-dir >/dev/null 2>&1; then \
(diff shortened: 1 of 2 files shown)


LukasWodka (changes requested):
Blocking this one too, same single cause as the rest of the fleet. Full analysis, the reproduction, the regression proof against `develop`, and a verified 3-line fix are on the canonical copy: **tracebloc/backend#2714**.

Short version: Bugbot's Medium about the parent-escape guard is correct. A `..` sitting **behind a missing component** — `core.hooksPath = missing/../../shared-hooks` — escapes, because the existence walk strips segments lexically and throws the `..` away along with the missing component, landing on `$PWD`, which reads as in-repo. `mkdir -p` then walks git's unresolved path physically, creates `missing/`, steps through `..`, and writes `pre-push` into the parent.

I reproduced it in a throwaway repo against **this PR's head**, with your Makefile and nothing else changed:

```
==> pre-push hook installed at missing/../../shared-hooks/pre-push
*** ESCAPED: wrote <parent>/shared-hooks/pre-push  (+ stray missing/ created in the repo)
```

And it's a regression: the same input against `develop` prints `outside this repo — skipping` and writes nothing. The old single `cd "$(dirname …)"` failed closed on a nonexistent component; the lexical walk that replaced it fails open.

The code is otherwise right and the guard block is byte-identical across all seven (re-verified at today's heads) — so the fix and the new test case apply here unchanged. Please take the patch and the extra case-4 entries from #2714 and I'll re-review straight away.

aptracebloc:
bugbot run

aptracebloc:
bugbot run

aptracebloc:
Fixed in the latest commit. You're right: the hook lives in `.git/hooks` and runs against whatever Makefile the pushed branch has, and `make` exits 2 for a **missing target** exactly as for a failed recipe — so `make guard-toolchain || exit 0` read "no such target" as "toolchain absent" and silently skipped `make check` on any branch predating this change.

The hook now probes existence first and only treats an *existing* target's failure as a skip signal:

```sh
if make -n guard-toolchain >/dev/null 2>&1; then make guard-toolchain >/dev/null 2>&1 || exit 0; fi
exec make check
```

A missing target falls through to `make check` (no silent skip); a present target with absent tools still soft-passes. Added a regression test — stubbed old Makefile (guard-toolchain has no rule) → `make check` still runs. Green under `sh` and `dash`, and applied byte-identically across all seven fleet PRs.

— drafted with Claude Code

LukasWodka (approved):
Approving — the missing-prefix escape is closed, and I verified it rather than reading the commit. Green: all checks pass, zero unresolved threads, Bugbot's Medium genuinely fixed rather than dismissed.

## The hole is closed, and the arm I worried about is intact

Same throwaway-repo repro as before, against this PR's head, plus the over-refusal arm in the same run:

```
escape (missing/../../shared)          -> contained
fresh-clone in-repo hooksPath, absent  -> installed-ok
```

Ran that across all seven heads: **contained + installed-ok on every one.** So the refusal didn't buy containment by breaking the fresh-clone case the ancestor-walk exists for.

## The new cases are mutation-proof, and I checked the anchor applied

Case 4c gained `missing/../..` and `missing/../../shared`; 4d is the not-yet-created in-repo control. I removed the `case "/$$sfx" in */../*) inr=0;; esac` line from the Makefile — asserting the anchor matched first, so an inert mutation couldn't pass for coverage — and re-ran the unchanged suite:

```
ok:   parent-escape refused: ..
ok:   parent-escape refused: ./..
FAIL: parent-escape refused: missing/../..            (got 'present', want 'absent')
FAIL: parent-escape refused: missing/../../shared     (got 'present', want 'absent')
ok:   in-repo but not-yet-created: hook installed
--- 2 FAILED ---
```

Exactly the right shape: the two new cases redden, the two old ones stay green because the earlier `dirname` fix already covered them, and 4d stays green because removing a refusal can't cause over-refusal. Full suite `ALL GREEN` on the six repos I could run standalone (backend's three failures are my synthetic repo lacking `.github/scripts/test_*.py`, not the change).

## Two things you did that I didn't ask for and would have missed

**Dropping `(resolves to '$$chd')` from the skip message.** With an unresolvable suffix, `$chd` is the deepest existing *ancestor*, not the target — so the old message would have printed the worktree root while explaining that the path is outside it. Removing it is the right call; keeping it would have been a small lie in the one line the developer reads.

**Widening 4c's assertion to `find` rather than a fixed `$esc/pre-push`.** The two new forms land in different places (`missing/../..` → `$esc`, `missing/../../shared` → `$esc/shared`), so a fixed path would have passed vacuously on the second one — present hook, absent assertion. And the `rm -rf missing shared` between forms stops one iteration's leftovers deciding the next. That's the failure mode a new case usually ships with.

Nice turnaround on this. The guard now refuses what it cannot resolve and installs what it can, and both halves have a case.

LukasWodka (commented):
**This merged with a regression, and my approval is why — flagging it here so it doesn't get lost.**

What happened, plainly: I verified this PR at head `a4b581cc`, `21aab12e` was pushed while I was checking the gate, and my approval landed on `21aab12e` — a commit I had not read. It merged at 15:13:33Z. My mistake: I should have re-read the head between verifying and approving, not assumed it was still the one I tested.

`21aab12e` is the right fix for a real Bugbot finding (`make` exits 2 for a missing target, so `make guard-toolchain || exit 0` read "no such target" as "toolchain absent" and skipped `make check` on older branches). But it reaches `exec make check` unconditionally when the probe fails, and the hook lives in `.git/hooks`, shared across branches. Against **`develop` as it stands now**:

```
hook installed from data-ingestors@develop
push from a branch with no Makefile  ->  exit 2  HARD-BLOCKED
    make: *** No rule to make target `check'.  Stop.
```

A Makefile present but without a `check` target does the same. The commit *before* it soft-passed both, so this is new — and it's the exact outcome the soft-pass exists to prevent, on clients that can't `--no-verify`.

One line fixes it, same `-n` trick applied to the target the hook actually invokes:

```make
	      'make -n check >/dev/null 2>&1 || exit 0' \
	      'if make -n guard-toolchain >/dev/null 2>&1; then make guard-toolchain >/dev/null 2>&1 || exit 0; fi' \
```

Measured across all four branch shapes — no Makefile, Makefile without `check`, Makefile with `check` but no `guard-toolchain`, modern Makefile with an unusable toolchain — and the full hook suite stays **21/21 ALL GREEN** including the new 9b. Strict strengthening.

Details and the sibling list are on tracebloc/backend#2714, where I've restated changes-requested; the other six are still open so the line can go in before they land. This repo needs a follow-up since it's already on `develop`.

Sorry for the noise — the fix here was good and I approved faster than I read.

PR #540, opened 2026-08-27
fix(devex): probe check in the pre-push hook (parity with the fleet)

aptracebloc:
## What

Brings data-ingestors' pre-push hook to parity with the rest of the backend#1749 fleet.

data-ingestors#538 (the fleet fix) landed one review round before the hook gained a `check` existence probe. Without it, a push from a branch that lacks `check` (or the Makefile) fails `make -n guard-toolchain`, falls through to `exec make check`, and **hard-blocks** on `No rule to make target 'check'` — with no `--no-verify` in the GUI clients the hook exists to protect (found by Lukas on backend#2714).

## Fix

```sh
make -n check >/dev/null 2>&1 || exit 0
if make -n guard-toolchain >/dev/null 2>&1; then make guard-toolchain >/dev/null 2>&1 || exit 0; fi
exec make check
```

The rule: the hook lives in `.git/hooks` and runs against whatever Makefile the pushed branch has, so it probes every target it invokes — it can outlive any of them. The `install-hooks` block is now byte-identical with the fleet's other Python repos.

## Tests

`make test-hooks` — adds case 9c (missing `check` target → soft-pass, no hard-block). Green under `sh` and `dash`.

Part of the fleet-wide fix for tracebloc/backend#1749.

— drafted with Claude Code

---

> [!NOTE]
> **Low Risk**
> Changes only local git hook install logic and its test script; no runtime, auth, or data-path impact.
> 
> **Overview**
> Aligns the **data-ingestors** pre-push hook with the rest of the backend#1749 fleet by probing **`make -n check`** before **`guard-toolchain`** and **`exec make check`**.
> 
> On branches (or trees) without a **`check`** target—or without a Makefile—the hook used to fall through and **hard-block** pushes with *No rule to make target 'check'*, which is especially painful in GUI git clients that often lack **`--no-verify`**. The new probe **soft-passes** (exit 0) when **`check`** is absent, matching the existing behavior for missing toolchain guards.
> 
> **`make install-hooks`** comments now state that every invoked target must be probed because the hook outlives any single Makefile revision. **`scripts/tests/test-pre-push-hook.sh`** adds case **9c** (missing **`check`** → soft-pass, **`make check`** not run) and tightens the **9b** stub **`make`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94ab53dad1383c7b18f5c94a77a8317b0b7c140b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,11 +243,15 @@ install-hooks:
 	      '# the hook cannot drift from it. guard-toolchain probes the TOOLS check' \
 	      '# runs, not installed packages (a broken venv still fails make check in a shell).' \
 	      '#' \
-	      '# guard-toolchain may be ABSENT on an older branch: this hook lives in' \
-	      '# .git/hooks and runs against whatever Makefile the pushed branch has, and' \
-	      '# make exits 2 for a MISSING TARGET just as for a failed recipe. Probe with' \
-	      '# -n so a missing target falls through to make check instead of silently' \
-	      '# skipping it (Bugbot, backend#1749).' \
+	      '# This hook lives in .git/hooks and runs against whatever Makefile the' \
+	      '# pushed branch has, so it can OUTLIVE any target it names — an older' \
+	      '# branch, staging before the train lands, or no Makefile at all. make' \
+	      '# exits 2 for a MISSING TARGET exactly as for a failed recipe, so probe' \
+	      '# every target with -n: an absent check (or Makefile) soft-passes rather' \
+	      '# than hard-blocking a push that has no --no-verify, and an absent' \
+	      '# guard-toolchain falls through to make check instead of skipping it' \
+	      '# silently (Bugbot + Lukas, backend#1749).' \
+	      'make -n check >/dev/null 2>&1 || exit 0' \
 	      'if make -n guard-toolchain >/dev/null 2>&1; then make guard-toolchain >/dev/null 2>&1 || exit 0; fi' \
 	      'exec make check' > "$$hook" && \
 	    chmod +x "$$hook" && \
--- a/scripts/tests/test-pre-push-hook.sh
+++ b/scripts/tests/test-pre-push-hook.sh
@@ -159,5 +159,17 @@ printf 'refs/heads/x deadbeef refs/heads/x 000\n' | env PATH="$stub:$PATH" sh "$
 check "$([ -f "$sent" ] && echo ran || echo skipped)" "ran" "old Makefile (no guard-toolchain) still runs make check"
 rm -rf "$stub"; rm -f "$sent"
 
+# 9c) an even older branch may lack `check` itself, or the Makefile entirely.
+# `make -n guard-toolchain` fails there too, so control would reach `exec make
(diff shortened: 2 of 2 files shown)


aptracebloc:
bugbot run

aptracebloc:
Good catch, fixed. Once the hook gained the `make -n check` probe, 9b's stub stamped its sentinel on *any* `check` argument — the dry-run probe included — so the "make check ran" assertion passed before `exec make check` was reached. 9b had quietly become a tautology.

Matched test 7's idiom: stamp only when `$1` = check (a real `make check`, never `make -n check`); guard-toolchain still reports no-rule. 9b is mutation-proof again — reverting the probe fix reddens 9b alone. Green under `sh` and `dash`, applied across all seven round-4 suites.

— drafted with Claude Code

LukasWodka (approved):
Approving — thanks for opening this yourself rather than leaving it to the follow-up I'd filed. Green, no open threads, and I verified the head I tested is the head I'm approving (a Makefile byte-compare against `94ab53da`, since approving a head I hadn't read is exactly what put this regression on `develop` in the first place).

## Behaviour, across the whole branch-shape matrix

Real hook installed from this head, real `make`, one ref line on stdin:

```
no Makefile at all                     -> exit 0   soft-passed
Makefile without `check`               -> exit 0   soft-passed
Makefile with `check`, no guard-toolchain -> make check RAN     (9b's case, intact)
modern Makefile, unusable toolchain    -> exit 0   soft-passed
```

The hard-block is gone and the older-branch case 9b exists for still runs.

## The parity claim is true

You wrote *"the `install-hooks` block is now byte-identical with the fleet's other Python repos."* Extracted and hashed it across the fleet:

```
data-ingestors        532fd470be23
tracebloc-py-package  532fd470be23     <- exact match
model-zoo             f7bd40025fd0     \
tracebloc-engine      f7bd40025fd0      |  differ from the above by ONE trailing
backend               f7bd40025fd0      |  blank line, nothing else
client-runtime        f7bd40025fd0     /
```

Diffed the two groups to be sure it wasn't hiding a logic difference — one blank line, no code. So the claim holds in the sense that matters.

## 9c is mutation-proof

Removed the `make -n check` line (anchor asserted first), left the tests untouched:

```
ok:   old Makefile (no guard-toolchain) still runs make check     <- 9b, still green
FAIL: missing check target soft-passes (exit 0, no hard-block) (got '2', want '0')
FAIL: soft-pass does not run make check (got 'ran', want 'skipped')
```

Exactly 9c reddens, 9b doesn't — the two cases test different things rather than the same thing twice.

**And I checked the `dash` half of your test claim rather than taking it**, since "green under `sh` and `dash`" is the kind of line that's easy to write and easy to have not run: `dash scripts/tests/test-pre-push-hook.sh` → **ALL GREEN** here too.

That closes the fleet. Sorry again for the lap — the hard-block was mine to catch before I approved #538, not after.

PR #546, opened 2026-08-28
fix(ingest-summary): carry an explicit record_count so token_classification stops inflating it (backend#2770)

aptracebloc:
## What & why

The ingest summary ships **no explicit row count**, so the backend infers a dataset's record count from `sum(labels.values())` — correct only when `labels` partitions the rows. Since data-ingestors#527, `token_classification`'s `labels` are per-**tag** occurrence counts (`get_tag_counts`), so that sum became total tokens ≈ avg sequence length × rows, silently inflating `total` / `unique_data_items_count`. #527 was right — it moved the error off the field that gates linking onto the field that doesn't — but the inflated count is now written for every token_classification dataset (re-)ingested after 2026-08-25.

This carries the count **explicitly** so the backend stops inferring it. Addresses the data-ingestors half of backend#2770.

## Change

- **`api/client.py`** — `send_ingest_summary` gains `record_count: Optional[int] = None`, added to the payload only when set (mirrors the `physical_table` conditional). Unset ⇒ payload byte-identical to today's, so a backend predating the field is unaffected — the two repos ship in either order.
- **`ingestors/base.py`** — passes `record_count` at the call site.

## ⚠️ One deliberate deviation from the issue text — please sanity-check

backend#2770 says "pass `stats["inserted_records"]`". Taken literally that **regresses `time_series_classification`**: it's sequence-grouped (`count_unit="sequences"`), so its `total` is deliberately the **sequence** count (backend#1054) — `inserted_records` is the *row* count (~mean(sequence length)× larger). Sending rows there and having the backend prefer it would re-inflate TSC's `total` exactly as the tag counts did for token_classification — the same bug one category over.

So this sends the category's **SAMPLE UNIT**, not a blanket row count:

| category | `record_count` |
|---|---|
| `time_series_classification` (grouped) | `number_of_sequences` |
| `token_classification` & every per-row category | `inserted_records` (rows) |

This is faithful to #2770's *intent* and fully compatible with the backend's `total = record_count or sum(labels.values())`: for every non-token_classification category `record_count` equals today's `total`, so only token_classification's number moves.

## Tests

- `token_classification`: `record_count` = **rows** while `labels` sum to **tokens** — the fixture where the two diverge (the whole regression).
- **TSC guard**: `record_count` = **3 sequences**, not the **15 stored rows** — pins the deviation above.
- Per-row control: `record_count` = rows.
- Client payload: field present only when set; absent (byte-identical) when unset.

## Verification (local)

- `ruff==0.15.20 check .` → clean · `black --check` (touched files) → clean
- `pytest tests/` → **2191 passed, 1 xfailed, 1 failed**. The one failure is `test_contract_consumers.py::…can_read_us`, **pre-existing & environmental** — it runs only against a locally checked-out `e2e-test-agent` that is stale (`SUPPORTED_VERSIONS={"2"}`) while `develop` already publishes layout `"3"`; this diff touches no schema/layout files, and CI uses a fresh consumer.

## Deploy / release

No lockstep with the backend PR (unknown field is ignored). Merging to `develop` publishes only a pre-release Python package — the signed **image** ships on a `v*` tag, so nothing auto-rolls to clusters on merge. It **unblocks the pre-2026-08-25 customer sweep** on backend#2662 only once an image carrying it reaches the fleet.

## Refs

**Part of tracebloc/backend#2770** — this is its data-ingestors half; the backend change (the `record_count` reader in `create_dataset_from_summary` / `combine_datasets`) is separate, so this PR does not close the ticket. Also: data-ingestors#527 (surfaced it) · backend#2662 / #1747 · not a blocker for backend#2660 / #1499.

### DIFF
--- a/tests/test_api_client_methods.py
+++ b/tests/test_api_client_methods.py
@@ -411,6 +411,23 @@ def test_summary_default_policy_sends_raw_labels():
     assert [s["label"] for s in sent["samples"]] == ["1", "0"]
 
 
+def test_summary_includes_record_count_when_set():
+    """backend#2770: an explicit record_count rides the payload as its own
+    field, independent of sum(labels.values()) — here labels sum to 30 while
+    the item count is 7."""
+    sent = _captured_summary_payload(_client(), record_count=7)
+    assert sent["record_count"] == 7
+    assert sum(sent["labels"].values()) == 30
+
+
+def test_summary_omits_record_count_when_unset():
+    """Default None omits the field, so a backend predating it is unaffected
+    and the payload stays byte-identical to today's (the two repos ship in
+    either order)."""
+    sent = _captured_summary_payload(_client())
+    assert "record_count" not in sent
+
+
 def test_summary_bucket_policy_buckets_labels_and_samples():
     from tracebloc_ingestor.utils.label_policy import BUCKET, apply
 
--- a/tests/test_time_series_classification.py
+++ b/tests/test_time_series_classification.py
@@ -260,6 +260,13 @@ def test_toy_csv_ingests_with_sequence_unit_counts(make_csv, monkeypatch):
     # number_of_columns still counts the CLEANED schema columns (k=2 site:
     # sequence_id + timestamp + F features; label stripped).
     assert summary_kwargs["meta_data"]["number_of_columns"] == 3
+    # backend#2770 (the TSC guard): record_count is the SAMPLE UNIT — the 3
+    # sequences, NOT the 15 stored rows. Passing inserted_records here would
+    # re-inflate total by ~mean(sequence length), the token_classification
+    # count bug (#527) one category over. This is the fixture where item count
+    # (3) and row count (15) diverge, so it pins the distinction.
+    assert summary_kwargs["record_count"] == 3
+    assert summary_kwargs["record_count"] != inserted  # 15 rows
 
 
 def test_toy_csv_requests_composite_index(make_csv):
@@ -285,8 +292,13 @@ def test_non_grouped_category_keeps_row_counts(make_csv):
     ing.database.get_label_counts.assert_called_once()
     ing.database.get_label_sequence_counts.assert_not_called()
     assert ing.database.create_table.call_args.kwargs["index_columns"] is None
-    meta = ing.api_client.send_ingest_summary.call_args.kwargs["meta_data"]
+    summary_kwargs = ing.api_client.send_ingest_summary.call_args.kwargs
+    meta = summary_kwargs["meta_data"]
     assert "number_of_sequences" not in meta
+    # backend#2770: a non-grouped category's record_count is the ROW count (15),
+    # the row unit like every non-grouped category — here it also equals
+    # sum(labels.values()), which is why the inference was ever right for them.
+    assert summary_kwargs["record_count"] == 15
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_token_classification_tag_counts.py
+++ b/tests/test_token_classification_tag_counts.py
@@ -150,6 +150,13 @@ def test_ingest_counts_distinct_tags_not_sequences():
     # Distinct TAGS (3), not distinct sequence strings (2).
     assert summary_kwargs["labels"] == {"O": 4, "B-PER": 2, "I-PER": 1}
     assert summary_kwargs["category"] == TOKEN
+    # backend#2770: the record count rides the payload EXPLICITLY and is the
+    # ROW count (2 texts) — NOT sum(labels.values()) (7 tag occurrences). This
+    # is exactly the fixture where the two diverge, which is the whole point:
+    # before this the backend inferred the count from the label sum, so #527's
+    # exploded tag counts silently inflated it by ~mean(sequence length)×.
+    assert summary_kwargs["record_count"] == 2
+    assert summary_kwargs["record_count"] != sum(summary_kwargs["labels"].values())
 
 
 def test_non_tag_category_keeps_row_counts():
--- a/tracebloc_ingestor/__init__.py
+++ b/tracebloc_ingestor/__init__.py
@@ -28,7 +28,7 @@ from .validators import (
 # Single source of truth for the package version. setup.py parses this literal
 # (see _read_version in setup.py) so the two can't drift again (#175). Bump here
 # only — setup.py picks it up automatically.
-__version__ = "0.8.14"
+__version__ = "0.8.15"
 
 __all__ = [
     "Config",
--- a/tracebloc_ingestor/api/client.py
+++ b/tracebloc_ingestor/api/client.py
@@ -223,6 +223,7 @@ class APIClient:
         meta_data: Optional[Dict[str, Any]] = None,
         physical_table: Optional[str] = None,
         label_policy: str = label_policy_module.PASSTHROUGH,
+        record_count: Optional[int] = None,
     ) -> Dict[str, Any]:
         """
         Send a single ingest summary to the backend, creating the UserDataSet in one
@@ -255,6 +256,16 @@ class APIClient:
                 (tracebloc/backend#1206). ``None`` (legacy shared-table
                 ingests) omits the field entirely, keeping the payload
                 byte-identical to today's.
+            record_count: The dataset's item count, carried EXPLICITLY so the
+                backend stops inferring it from ``sum(labels.values())``
+                (backend#2770). That inference is only right when ``labels``
+                partitions the rows; token_classification's exploded per-tag
+                counts (data-ingestors#527) and any sequence-grouped
+                category's per-sequence counts both break it. The caller
+                passes the category's SAMPLE UNIT (rows, or sequences for a
+                grouped category). ``None`` omits the field, so an older
+                backend that predates it is unaffected and the payload stays
+                byte-identical to today's — the two repos ship in either order.
 
         Returns:
             ``{"dataset_id": ..., "dataset_key": ...}``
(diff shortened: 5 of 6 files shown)


shujaatTracebloc (commented):
The reasoning here is right, including the part you flagged for a sanity-check. Commenting rather than approving only because the gate is red — two real failures, both mechanical, details at the bottom.

## The deviation from the issue text is correct — keep it

You asked for a sanity-check on sending the sample unit rather than `stats["inserted_records"]` blanket. I agree, and I'd go further: taking #2770 literally would have shipped the same bug one category over. `time_series_classification` is sequence-grouped, its `total` is deliberately the sequence count (backend#1054), so sending rows there and having the backend prefer them would inflate TSC's `total` by ~mean(sequence length) — structurally identical to what the exploded tag counts did to token_classification. The issue text was written from the token_classification case and didn't account for the grouped one. Your table is the faithful reading of its intent.

I verified the branch selection rather than taking the comment's word for it. The new condition at `base.py:1343` is byte-identical to the count-path condition at `:1215`, and `self.file_options["number_of_sequences"]` is assigned at `:1226` *inside that same branch* — same indentation level, same straight-line path through `_ingest_with_lock`, no intervening conditional. So the key is guaranteed present whenever the new code reads it; there's no `KeyError` on a grouped category. The "mirrors the count-path branch selection above" claim holds literally, which is what makes the two numbers consistent by construction rather than by coincidence.

Your read on the local `test_contract_consumers` failure is also correct — that's a stale local `e2e-test-agent` checkout. `SUPPORTED_VERSIONS` went to `{"2", "3"}` on its develop, so a fresh consumer is green. Nothing for you to do.

## One thing worth fixing while the backend half is still unwritten

`client.py` guards with `is not None` and the comment explains why: *"the guard keeps a legitimate 0 from silently dropping to the backend's sum-of-labels fallback."*

But #2770 specifies the backend as `total = summary.get("record_count") or sum(labels.values())` — truthiness. Under that, `record_count=0` **is** falsy and **does** drop to the fallback. The careful `is not None` on this side is defeated one hop later, so the comment promises a protection the contract doesn't deliver.

It isn't a live bug: a 0-row run doesn't reach the call site, and if it did, `labels` would be empty and the fallback would also yield `0` — same answer either way. So nothing misbehaves today. It's worth raising only because the backend half doesn't exist yet, which makes it free to fix: specifying that side as `record_count if record_count is not None else sum(...)` costs nothing, makes the two halves agree on the one value they currently disagree about, and means this comment stays true. As written, the asymmetry is latent — it only becomes real if the empty-labels fallback ever stops being `0`, and that's the kind of thing nobody re-derives years later.

Either align the backend spec, or soften the comment here to say the `0` case is unreachable rather than protected. I'd pick the first.

## The two failures

**`version-bump-gate`** — `tracebloc_ingestor/__init__.py` is still `0.8.14`, that version is already released, and this PR touches `tracebloc_ingestor/api/client.py` (matches the `tracebloc_ingestor/*` publish path). Bump it. Worth noting the gate's own reasoning: leaving it stale doesn't fail at merge, it fails the next prod hop days later on somebody else.

**`set-status / closing-ref`** — the title names `backend#2770` but nothing in the body links it, and a title reference is inert. Given this is only the data-ingestors half, `Closes` would be wrong — use:

```
Part of tracebloc/backend#2770
```

Fully qualified, since a bare `#2770` resolves against this repo. That satisfies the gate without promising a close. And you don't need a manual re-run — this repo now listens for `edited`, so saving the body re-fires the check on its own.

shujaatTracebloc (commented):
`version-bump-gate` is green — `0.8.14 → 0.8.15` in `06e89967` does it. `closing-ref` is still failing, and it's worth being precise about why, because the current body is one word-order away from a fix that would be actively wrong.

The Refs line reads:

> backend#2770 (this closes its data-ingestors half)

The keyword lands *after* the number. GitHub only creates a link from `<keyword> <ref>`, so that text links nothing and the gate still sees the title naming a ticket the body doesn't reference. The trap is what happens next: the obvious repair is to move the word in front — but `Closes tracebloc/backend#2770` would auto-close the ticket the moment this merges, and this is only the ingestor half. The backend serializer, `create_dataset_from_summary`, and the `combine_datasets` path are all still open, so the sweep on backend#2662 would be left blocked by a ticket that reads as done.

Your instinct not to promise a full close is right. Just make it a form that links:

```
Part of tracebloc/backend#2770
```

Fully qualified — a bare `#2770` resolves against this repo. That satisfies the gate, creates the reference, and promises nothing untrue. Saving the body re-fires the check on its own; no push needed.

Still holding on the gate (Bugbot pending). The substantive note from my last pass — `is not None` here versus `or` in #2770's backend spec — stands and is still free to fix while that half is unwritten.

aptracebloc:
Thanks @shujaatTracebloc — and for verifying the branch selection independently rather than trusting the comment; that's exactly the invariant that makes the two counts consistent by construction.

**The `is not None` vs `or` note — addressed, and I agree with your preferred option.**

- **This side (done, `56c6bbd`):** the comment no longer claims an end-to-end protection it can't deliver alone. It now states only what the ingestor guarantees — send the explicit value including a legitimate `0` — notes the `0` is unreachable in practice (a 0-row run is skipped upstream), and says plainly that for an explicit `0` to survive, the backend must read it with the same `is not None`, tracked on backend#2770.
- **Backend half (your option a — align the spec):** agreed, and it's free while that half is unwritten. The exact change to #2770's spec is `or` → explicit `is not None` in both places the count is derived:
  ```python
  # create_dataset_from_summary AND combine_datasets
  total = record_count if record_count is not None else sum(labels.values())
  ```
  That makes the two halves agree on the one value they currently disagree about, and keeps my `is not None` on this side honest end to end. Want me to update #2770's "The fix" bullet to that form, or would you rather the backend PR carry it?

**Gate failures both resolved:** `version-bump-gate` green (`0.8.14 → 0.8.15`, `06e8996`); `closing-ref` green — body now carries `Part of tracebloc/backend#2770` (keyword before the fully-qualified ref, so it links without promising a close, since the backend serializer / `create_dataset_from_summary` / `combine_datasets` are all still open).

All 15 checks green and both Bugbots clean as of `56c6bbd`. Re-requested you and @saadqbal since your review dropped the requests.

aptracebloc:
The backend half is up as tracebloc/backend#2774 and carries the alignment you preferred: `total = record_count if record_count is not None else sum(labels.values())` in **both** `create_dataset_from_summary` and `combine_datasets`, so an explicit `0` survives end to end and the two halves agree on the one value they disagreed about. No spec edit needed — the PR is the source of truth now.

The backend PR also propagates the count per-edge through `_merge_dataset_summary` so `combine_datasets` re-derives the combined total from the row counts, not the merged label sum.

— drafted with Claude Code

shujaatTracebloc (approved):
Approving — all checks green, no open threads.

Everything I raised is addressed, and the comment correction in `56c6bbd1` is the right call rather than an overcorrection. It keeps `is not None` on this side (correct — this side genuinely shouldn't drop a computed `0`) while being honest that the guard is one-sided until the backend reader uses the same test. That's more useful than either the original claim or silently switching to `or`, and pointing at backend#2770 puts the note where whoever writes that half will find it.

I checked the new factual claim rather than taking it on trust: `base.py:1270` bails with "skipping ingest summary" when `inserted_records` is falsy, well before the `send_ingest_summary` call at `:1347`. So the `0` really is unreachable in practice, exactly as the comment now says.

The Refs line is also right. `Part of tracebloc/backend#2770`, fully qualified, with an explicit note that the backend reader is separate — links the ticket, satisfies the gate, and doesn't promise a close this PR can't deliver.

Recapping what I verified on the substance, since this one crosses repos:

- The deviation from the issue text — sending the category's sample unit rather than a blanket `inserted_records` — is correct, and taking #2770 literally would have inflated `time_series_classification`'s `total` by ~mean(sequence length), the same bug one category over.
- The new branch at `base.py:1343` is byte-identical to the count-path condition at `:1215`, and `number_of_sequences` is assigned inside that same branch at `:1226` on the same straight-line path, so there's no `KeyError` on a grouped category and the two numbers agree by construction.
- Unset `record_count` leaves the payload byte-identical, so the two repos ship in either order.

One thing for whoever picks up the backend half: the PR body still describes it as `total = record_count or sum(labels.values())` from the issue text. Worth writing that side as `is not None` so the two halves agree on `0` — no behavioural difference today, but it makes the comment here true end to end.

Nice piece of work. Catching that the issue's own prescription would regress a neighbouring category is the part that mattered.

saadqbal (approved):
Approving — and the deviation from the issue text is the right call. Sending `inserted_records` for a sequence-grouped category would have made the backend prefer rows over sequences and re-inflated TSC's `total` by mean sequence length, i.e. the same bug one category over. Good catch, and the guard test pins it (I mutated the branch away and it reddens; so does swapping in `sum(label_counts.values())`, and dropping the payload field reddens the client test).

One caveat on "every non-token_classification category's `record_count` equals today's `total`": today's number is `sum(label_counts.values())`, which is a `COUNT(*)` off the stored rows, while `inserted_records` is accumulated from `insert_batch`'s `success_ids` — and those diverge when a `data_id` repeats across batches, since the upsert collapses the row but each batch's SELECT re-counts it. The objdet content_hash case you already warn about at `base.py:294` is exactly that shape. Narrow and in the same direction as the bug being fixed, so not worth holding the PR, but the claim is "equal except for duplicate-collapsing datasets", and a `COUNT(*)`-by-ingestor_id helper would make it unconditionally true. Worth a line on backend#2770 or a follow-up.

No other reader of the field in this repo, `number_of_sequences` in `meta_data` agrees with `record_count` for the grouped path rather than contradicting it, and the field name matches what the backend half is specced to read.

PR #547, opened 2026-08-28
release-train: develop -> staging

LukasWodka:
Automated promotion by the release train (RFC-0008 D14). Head is the train-managed `release-train/to-staging` branch (a mirror of `develop`), so it never collides with a human PR. Merged only when the fr-gate is green. 

---

> [!NOTE]
> **Medium Risk**
> Changes registration metadata sent to the backend for every successful ingest; incorrect counts would misreport dataset size, but logic is category-scoped and backward-compatible when the field is omitted.
> 
> **Overview**
> Adds an optional **`record_count`** field to the ingest-summary API payload so the backend no longer infers dataset size from **`sum(labels.values())`**, which is wrong for **token classification** (per-tag label sums) and **sequence-grouped** tasks like time-series classification (per-sequence labels vs stored rows).
> 
> **`BaseIngestor`** computes the count in the category’s sample unit: **sequences** when grouping uses `count_unit == "sequences"`, otherwise **inserted row count**, and passes it into **`send_ingest_summary`**. **`APIClient`** includes **`record_count`** in JSON only when the value is not `None`, keeping older backends byte-compatible when the field is omitted.
> 
> Tests lock in payload shape (present vs absent) and category-specific behavior (TSC: 3 sequences vs 15 rows; token classification: 2 texts vs 7 tag occurrences; tabular: row count). Package version **0.8.14 → 0.8.15**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7235c955814a03267f923bfef9e9260eea8b5185. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_api_client_methods.py
+++ b/tests/test_api_client_methods.py
@@ -411,6 +411,23 @@ def test_summary_default_policy_sends_raw_labels():
     assert [s["label"] for s in sent["samples"]] == ["1", "0"]
 
 
+def test_summary_includes_record_count_when_set():
+    """backend#2770: an explicit record_count rides the payload as its own
+    field, independent of sum(labels.values()) — here labels sum to 30 while
+    the item count is 7."""
+    sent = _captured_summary_payload(_client(), record_count=7)
+    assert sent["record_count"] == 7
+    assert sum(sent["labels"].values()) == 30
+
+
+def test_summary_omits_record_count_when_unset():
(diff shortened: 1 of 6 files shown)


LukasWodka:
bugbot run

PR #548, opened 2026-08-28
release-train: staging -> main

LukasWodka:
Automated promotion by the release train (RFC-0008 D14). Head is the train-managed `release-train/to-main` branch (a mirror of `staging`), so it never collides with a human PR. Merged only when the fr-gate is green. 

---

> [!NOTE]
> **Medium Risk**
> Changes how dataset totals are reported to the backend at registration time; wrong counts would affect downstream metadata until the backend half of #2770 is aligned. The workflow change is low-risk CI behavior.
> 
> **Overview**
> **Ingest summary (backend#2770)** — The ingestor now sends an optional **`record_count`** on the global-meta summary payload so the backend does not infer dataset size from **`sum(labels.values())`**. That sum is wrong for **token classification** (per-tag label counts vs row/text count) and **time series classification** (per-sequence labels vs stored rows). **`BaseIngestor`** sets the count to the category’s sample unit: **sequences** when grouping uses `count_unit == "sequences"`, otherwise **inserted row count**. **`APIClient.send_ingest_summary`** adds the field only when the caller passes a non-`None` value, keeping older backends unchanged. Package version **0.8.15**.
> 
> **CI** — **`set-pr-status.yml`** listens for **`pull_request` `edited`** so title/body changes re-run the closing-ref gate (same fix as elsewhere; avoids bypass and stuck red status after fixes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a9f9768db2d3e1424743f2149efc3b7b1370f2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/set-pr-status.yml
+++ b/.github/workflows/set-pr-status.yml
@@ -2,7 +2,25 @@ name: Set PR Status
 
 on:
   pull_request:
-    types: [opened, reopened, ready_for_review, converted_to_draft]
+    # `edited` is load-bearing, not decoration (tracebloc/backend#2556). The
+    # `closing-ref` job decides its verdict from the PR TITLE and the PR BODY, and
+    # `edited` is the ONLY event GitHub fires when either changes. Without it the
+    # two fields the gate reads are the two fields that can change without
+    # re-running it, which is a bypass rather than a gap: open a PR titled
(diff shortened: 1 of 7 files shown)


LukasWodka:
bugbot run

